PR #18, opened 2016-09-21
Fix gulp tslint and change new number field default value

harunurhan:


harunurhan:
@jmartinm I forgot about this and I need the fix in the first commit, so I will include this in the large upgrade PR or will do it later. That's why closing now.

PR #27, opened 2016-09-30
remove save button and save logic (fix #15)

### LINKED ISSUE #15: Remove "Save" button.
I think it should be user's responsibly to implement the logic of save button. Since edited `json` document is always accessible from  `EditorComponent`.

Advantage of this is that the editor would be more flexible.

Disadvantage of this is that we should also remove `x_remote_validation_url` which is being in `click` event of the button. However we can have an option input to editor like `errors` so that we user can do remote validation and passed to errors to the editor to be displayed in the UI.

In case we decide to remove, I can move the `Save` button to example app.

harunurhan:
* Removes save-button and x_editor_validation_url schema extension.

* Adds errorMap input to json-editor so that errors can be passed
to be showed on the UI.

* Updates README.md.


README.md
```diff
@@ -26,6 +26,11 @@ Then add it to your module's `imports`.
 <json-editor [schema]="mySchema" [record]="myRecord" (onRecordChange)="doStuffWithNewRecord($event)"></json-editor>
 ```
 
+- `schema` : valid json-schema for the record.
+- `record` : valid json to be edited.
+- `onRecordChange` emitted when record change, `$event` is the edited record.
```
harunurhan: It has to be called event `$event` in `html`, because it something that is provided by `angular2`. I just renamed the ones that are in `typescript`.

example/app/app.component.ts
```diff
@@ -24,6 +24,8 @@ import { Component, ViewEncapsulation } from '@angular/core';
 import { Http } from '@angular/http';
 import { Observable } from 'rxjs/Observable';
 import 'rxjs/add/observable/zip';
+import 'rxjs/add/operator/map';
```
harunurhan: good catch!, actually no. `tslint` actually complains about unused imports but apparently, it doesn't check `example` folder.

Issue #37, opened 2016-10-06
Return clean json from json-editor


harunurhan:
Currently we apply some fixes on the record to make UI (usually `table-list`) to look cool. But this changes structure of the record (flattening) and inserts empty values.

To solve this, we'll remove flattening and instead of inserting empty values when a key undefined, just use a pipe or some function like `selfOrEmptyValueIfUndefined` on the template so that we could fix the UI without changing the record. So we won't need to run a cleanup function

harunurhan:
I realized that we also do some manipulations for `x_editor_hidden` and `x_editor_always_show`, so do we have to deal with them unfortunately.

My idea for `x_editor_always_show` is that we won't need to force it on input record but only when a new value added. And for `x_editor_hidden` we can implement functionality that passes those fields in `mapToSortedIterable` pipe or in a new pipe.

what do you think @jmartinm ?

jmartinm:
> My idea for x_editor_always_show is that we won't need to force it on input record but only when a new value added

This one I am not sure. The idea with `x_editor_always_show` is that when you open the record those fields are shown in the interface to guide what fields are recommended to be filled.

So it is not only useful with new value added but also upon opening the record.

jmartinm:
Done in https://github.com/inveniosoftware-contrib/ng2-json-editor/pull/54.

PR #38, opened 2016-10-13
add any-type-field, increase json type coverage

harunurhan:
* Adds any-type-field component to share template code which decides
what component to use according to the schema.

* Increases json type coverage such as object in an object, array in
array

* Move x_editor_hidden handling into mapToSortedIterable pipe.
(referencing #37)

* Removes flattening/unflattening (referencing #37).

* Updates mock-data to latest version schema and record of Inspire.

### DIFF
--- a/src/shared/services/component-type.service.ts
+++ b/src/shared/services/component-type.service.ts
@@ -38,6 +38,10 @@ export class ComponentTypeService {
    * @return {string}
    */
   getComponentType(schema: Object): string {
+    if (!schema) {
+      throw new Error('schema is undefined');
+    }
+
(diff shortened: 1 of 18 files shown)


src/shared/services/component-type.service.ts
```diff
@@ -48,7 +52,13 @@ export class ComponentTypeService {
 
     let schemaType = schema['type'];
     if (!schemaType) {
-      return 'raw';
+      if (schema['anyOf']) {
+        return 'any-of';
```
harunurhan: This is not to crash the editor when editor, so we can merge this as well.

PR #51, opened 2016-10-18
layout and css improvements

### LINKED ISSUE #47: UI improvements
Current main pain points are:

* Too much white space in between fields.

* Add field button might be better below the field name.

* Clean up non-editable fields. For example curated_relation should not be visible. In the future that field can be used to give a hint to the cataloger that the field was already edited manually. It should be automatically set to True when the field is edited (connected to https://github.com/inspirehep/record-editor/issues/39)

* Authors and references have nested structure that will need a special display. Some parts might be folded by default. The nested structure with tables inside of tables makes it hard to read in what field we are on.

* The page should have fixed height with Preview and Record being scrollable, but not the whole page.

* We need space to add the actions (from https://github.com/inveniosoftware-contrib/invenio-record-editor/issues/2)

jmartinm:
* Addresses #47.

### DIFF
--- a/src/add-field-dropdown/add-field-dropdown.component.scss
+++ b/src/add-field-dropdown/add-field-dropdown.component.scss
@@ -1,4 +1,5 @@
 .editor-btn-add-subfield {
+  padding: 0;
   font-size: 21px;
   opacity: 0.4;
   border: 0;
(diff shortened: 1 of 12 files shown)


jmartinm:
Just a start to try to make the layout better.

editor

harunurhan:
LGTM 👍

PR #53, opened 2016-10-20
enrich schema with configuration

jmartinm:

### DIFF
--- a/src/shared/services/schema-fixer.service.ts
+++ b/src/shared/services/schema-fixer.service.ts
@@ -20,21 +20,27 @@
  * as an Intergovernmental Organization or submit itself to any jurisdiction.
 */
 
-declare var System: any;
+import { Injectable } from '@angular/core';
 
-// removes the error when testing!
-interface AutocompletionResult {
-  text: string;
-  payload?: Object;
-}
+import { JsonUtilService } from './json-util.service';
 
-interface AutocompletionOptions {
-  url: string;
-  path: string;
-  size: number;
-}
+@Injectable()
+export class SchemaFixerService {
 
-interface AppConfig {
-  schemaOptions?: Object;
-  previews?: Array<Object>;
+  constructor(private jsonUtilService: JsonUtilService) { }
+
+  /**
+   * Enriches schema with options coming from configuration
+   * 
+   * @param {Object} schema - schema for the record
+   * @param {Object} config - schema specific options
+   * @return {Object} - fixed record
+   */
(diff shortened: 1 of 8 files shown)


src/shared/services/schema-fixer.service.ts
```diff
@@ -0,0 +1,50 @@
...
+  /**
+   * Enriches schema with options coming from configuration
+   * 
+   * @param {Object} schema - schema for the record
+   * @param {Object} config - schema specific options
+   * @return {Object} - fixed record
+   */
+  fixSchema(schema: Object, config: Object): Object {
+    Object.keys(config).forEach(field => {
+      let schema_field = this.jsonUtilService.getValueInPath(schema, field);
```
harunurhan: not so important but I generally use `schemaField` naming convention.

src/typings.d.ts
```diff
@@ -31,3 +31,7 @@ interface AutocompletionOptions {
   path: string;
   size: number;
 }
+
+interface AppConfig {
```
harunurhan: since this config is for editor, it might be better to name it here `EditorConfig`

Actually I usually add it next to declaration when it has only single property like

`@Input config: {schemaOptions?:  Object};`

jmartinm: Was thinking that at the moment the config is for jsonschema properties, but we might have some other config to drive the behaviour of the editor (and not tight to a field path) so we will expand this interface.

jmartinm:
Was updating as well the README (see https://github.com/jmartinm/ng2-json-editor/blob/45b54d192a3ed2e35b7fa3cb49446701d1806f5f/README.md#schemaOptions)

At the moment all the options are related to a given field in the jsonschema except `x_editor_previews`  (that needs to be in top level given as a list) right? Was there a reason for that or shall we configure the previews in the same way?


```
{
  'titles.0.title': {
    x_editor_hidden: true
  },
  'abstracts': {
    x_editor_preview: {
        {
             name: string; (name of the tab which preview is displayed)
             type: string; (type of the preview, such as "pdf")
        }
  }
}
```

harunurhan:
@jmartinm Since it needs to be extracted at the beginning I thought it should in the top-level. Also it is not part of any field UI.

jmartinm:
> @jmartinm Since it needs to be extracted at the beginning I thought it should in the top-level. Also it is not part of any field UI.

Makes sense, then we might want to add it as another configuration, `schemaOptions` and `previews` for the config introduced in this PR.

PR #54, opened 2016-10-20
Make json clean again

harunurhan:


src/complex-list-field/complex-list-field.component.ts
```diff
@@ -36,16 +36,66 @@ import {
...
+
+    this.keys = this.values
+      .map(value => Object.keys(value));
+  }
+
+  onFieldAdd(field: string, index: number) {
+    this.keys[index].push(field);
+  }
+
+  /**
```
harunurhan: So below, you can see that I overridden some methods to update keys as well when values change. But instead we could put `ngOnInit` logic to `ngOnChange` in order to avoid below code.
@jmartinm

jmartinm: 👍 I think it can make sense, was there a reason not to do it? You think it will be heavy to do the `.map` every time? (might be with big lists of references)

src/tree-menu/tree-menu-item.component.ts
```diff
@@ -49,6 +53,17 @@ export class TreeMenuItemComponent extends AbstractTrackerComponent implements O
 
   ngOnInit() {
     this.href = `${this.windowHrefService.getHrefWithoutHash()}#${this.path}`;
+
+    if (this.schema['type'] === 'object') {
+      this.keys = Object.keys(this.value);
+    }
+  }
+
+  ngOnChanges(changes: SimpleChanges) {
```
jmartinm: I see that the tree map only seems to realise about top level changes. So if there is author 0 and I add a new one, author 1 appears in the tree map.

But if I add fields to either author, these don't appear in the tree map HTML.

jmartinm: I see that the tree map only seems to realise about top level changes. So if there is author 0 and I add a new one, author 1 appears in the tree map.

But if I had fields to either author, these don't appear in the tree map HTML.

jmartinm:
Will merge this to base new changes on top of this PR and we can see if something can be improved later on. (will create an issue for the tree map)

Issue #55, opened 2016-10-21
tree map reflect changes in complex objects


jmartinm:
The tree map does not update when changes happen deep into complex objects.

![editor](https://cloud.githubusercontent.com/assets/580727/19595593/ecd1dd5e-978a-11e6-84a2-a68613b893bb.gif)

PR #56, opened 2016-10-24
WIP: record change notifier

jmartinm:


jmartinm:
This is something I was trying but it is not working yet. The idea was to have a shared way of notifying changes so that components can act on it (to solve for example https://github.com/inveniosoftware-contrib/ng2-json-editor/issues/55).

Some problems at first sight:

* When one value changes, the subscription runs many times https://github.com/inveniosoftware-contrib/ng2-json-editor/pull/56/files#diff-0ddd2c1b9e29bb90fd5ff601c886fd5aR54

* The tree menu is throwing an error `Cannot read property 'x_editor_hidden' of undefined` I guess because I might be putting something wrong in the keys that do not exist in the schema.

So just a small start so that we can discuss here @harunurhan

jmartinm:
Not needed at the moment.

PR #57, opened 2016-10-26
create add-new-element-button component

harunurhan:
- Creates add-new-element-button component that enables to move
  add-new-element button and its logic to the parent so that it will
  appear in the space under the key/title of property.

### DIFF
--- a/src/complex-list-field/complex-list-field.component.ts
+++ b/src/complex-list-field/complex-list-field.component.ts
@@ -76,14 +76,8 @@ export class ComplexListFieldComponent extends AbstractListFieldComponent implem
     this.keys[newIndex] = temp;
   }
 
-
-  /**
-   * @override
-   * to update keys
-   */
-  addNewElement() {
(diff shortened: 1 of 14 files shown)


src/complex-list-field/complex-list-field.component.ts
```diff
@@ -76,14 +76,8 @@ export class ComplexListFieldComponent extends AbstractListFieldComponent implem
...
-   * @override
-   * to update keys
-   */
-  addNewElement() {
-    super.addNewElement();
-
+  onNewElementAddToChildsProperty(value: Array<any>, index: number, key: string) {
+    super.onValueChange(value, index, key);
     // update keys array when element moved
     this.keys.push([]);
```
jmartinm: This does not belong in here.

harunurhan: This will be fixed by the immutable PR since we need immutables to fix it.

PR #64, opened 2016-11-02
add limited support for anyOf

harunurhan:
* Adds anyOf support where all anyOf items have same properties
and are objects. Picks the first one, merges all enum values and
removes patter and format rules.

* Adds tests for schema-fixer service

Shall we add the info about limitation of this support (like in the commit message) to somewhere
in README?
@jmartinm

### DIFF
--- a/src/shared/services/schema-fixer.service.ts
+++ b/src/shared/services/schema-fixer.service.ts
@@ -30,17 +30,86 @@ export class SchemaFixerService {
   constructor(private jsonUtilService: JsonUtilService) { }
 
   /**
-   * Enriches schema with options coming from configuration
+   * Fixes schema to be in a format that expected by json-editor
    * 
-   * @param {Object} schema - schema for the record
+   * @param {Object} schema - json schema
    * @param {Object} config - schema specific options
-   * @return {Object} - fixed record
+   * @return {Object} - fixed schema
    */
   fixSchema(schema: Object, config: Object): Object {
+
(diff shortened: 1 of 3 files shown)


src/shared/services/schema-fixer.service.ts
```diff
@@ -30,17 +30,86 @@ export class SchemaFixerService {
...
+          enums[propName] = [];
+        }
+        enums[propName] = enums[propName].concat(anyOfElement['properties'][propName]['enum']);
+      });
+    });
+    let fixedSchema = anyOf[0];
+    // shallow cleaning of format and pattern
+    Object.keys(fixedSchema['properties'])
+      .forEach(prop => {
+        delete fixedSchema['properties'][prop]['format'];
```
jmartinm: Shall we delete specific ones or filter the ones we would like to keep (like `value`) and delete the rest? I am saying in case there is some other property that could have unexpected side effects.

jmartinm:
> Shall we add the info about limitation of this support (like in the commit message) to somewhere
in README?

yes, would be good to add it in the [use](https://github.com/inveniosoftware-contrib/ng2-json-editor/blob/master/README.md#use) section. Where you say `schema : valid json-schema for the record.` There can be a link `See json schema limitations` that points to a new section where we document all the limitations we encounter (like this anyOf)

jmartinm:
Just tried locally and seems to work ok.

PR #65, opened 2016-11-03
Searchable dropdown fix

### LINKED ISSUE #43: searchable-dropdown should select when pressed enter
Currently it has a problem, it doesn't select the focused dropdown item.

Also, if there is a single options, `enter` should select it even if it is not on focus or it can always focus the first item which will directly support this functionality when the above bug is resolved.

### LINKED ISSUE #44: x_editor_enum_shortcut_map should be documented on the README

harunurhan:
### refactor and fix searchable-dropdown

* Changes from prefix search to ignored case expression search which tries to
match given input in anywhere of possible items.

* Fixes #43

### add documentation for x_editor_enum_shortcut_map

* Fixes #44

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,6 +123,19 @@ number
 default: 0
 ```
 
+#### x_editor_enum_shortcut_map
+
+Shortcut map for enum fields, to map shorthand values to enum values
+
+```
+{
+  a: "An value in enum array",
+  abc: "Another value in enum array"
+}
+```
+
+So when `a` typed in enum field and enter is pressed, enum field value will be set to `Another value in enum array`.
+
 ### <a name="previews"></a>Previews
 
 Configuration for previews to be displayed in previewer (on the right side).
--- a/src/shared/pipes/filter-by-expression.pipe.spec.ts
+++ b/src/shared/pipes/filter-by-expression.pipe.spec.ts
@@ -20,16 +20,21 @@
  * as an Intergovernmental Organization or submit itself to any jurisdiction.
 */
 
-import { Pipe, PipeTransform } from '@angular/core';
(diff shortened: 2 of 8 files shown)


README.md
```diff
@@ -123,6 +123,19 @@ number
 default: 0
 ```
 
+#### x_editor_enum_shortcut_map
+
+Shortcut map for enum fields, to map shorthand values to enum values
+
+```
+{
+  a: "An value in enum array",
```
jmartinm: This is a `shortcut` and not value in enum array?

README.md
```diff
@@ -123,6 +123,19 @@ number
...
+  abc: "Another value in enum array"
+}
+```
+
+So when `a` typed in enum field and enter is pressed, enum field value will be set to `An value in enum array`.
```
jmartinm: The enum field value will be `Another value in enum array` ?

src/shared/pipes/filter-by-expression.pipe.spec.ts
```diff
@@ -20,17 +20,17 @@
...
 
   beforeEach(() => {
-    pipe = new FilterByPrefixPipe();
+    pipe = new FilterByExpressionPipe();
   });
 
-  it('should filter by prefix', () => {
-    let values = ['ab1', 'ab2', 'ac1', 'bc1'];
+  it('should filter by expression', () => {
+    let values = ['ab1', 'ab2', 'xc1', 'zc1'];
```
jmartinm: Would be good to test that your regexp filtering works, for example typing c1 should return 'xc1', 'zc1'

README.md
harunurhan: key is `shortcut`, value should be one of the values in enum array.

Issue #45, opened 2016-10-14
Support for custom components for particular fields


harunurhan:
It should support custom `angular2` components for particular fields.

harunurhan:
Custom template support for particular fields, we can do something like this [typeahead directive](http://valor-software.com/ng2-bootstrap/#/typeahead). 

We should have configuration which is a function like:
```javascript
function template(value) {
  return `<div>valid html string that makes use of ${value}<div>`;
}
```

Problem with this, this could only be useful for custom components of primitive fields.

PR #71, opened 2016-11-08
implement customizable $ref field

harunurhan:
* Adds dynamic-templater-loader service that can compile and load angular2 templates
into DOM in runtime by wrapping them with a dummy component and a
module.(referencing #45)

* Adds ref-field component that handles the object field with $ref
inside using the template in x_editor_ref_config from schema.


src/ref-field/ref-field.component.html
```diff
@@ -0,0 +1,8 @@
+<div [id]="path" [ngSwitch]="shouldDisplayWithTemplate" >
+  <div *ngSwitchCase="true">
+    <button class="btn-preview-template" *ngIf="isLazy" (click)="onCheckClick($event)">👁</button>
```
jmartinm: small thing, but `onPreviewClick` might be better name.

jmartinm:
LGTM.

PR #83, opened 2016-11-18
On change cb config

### LINKED ISSUE #75: object-field delete doesn't work

harunurhan:
#### add global store for top level json input
* Adds json-store service to read from/write to top level json input
via path which removes the @Output and event emitting structure and
make it possible to be read/written any part of json by any component.

* Fixes tree-menu focus first input element issue caused by filter
input in add-field-dropdown.

* Fixes #75

#### add on value change and completion select config
* Adds `x_editor_on_value_change` config that takes a function
with params like value, path, store (JsonStoreService/NestedStore) to be
executed when configured field is changed. It can do any manipulation
on top level json input.

* Adds NestedStore type to be able use, JsonStoreService type in
typings because importing something into typings.d.ts breaks it.

* Adds `onCompletionSelect` to `x_editor_autcomplete` that is executed
when a completion from dropdown is selected and it can access all
record via store (JsonStoreService/NestedStore).

* Avoids calling the remote autocompletion service when the input is
empty.

* Updates README accordingly.

* Fixes autocomplete-input component test

#### optimize ref field and add headers option
* Optimizes ref-field dynamic template loader by changing the context
of dynamic template instead of create a new one everytime.

* Adds headers option to ref config.

* Fixes minor bugs in ref-field.

### DIFF
--- a/src/ref-field/ref-field.component.html
+++ b/src/ref-field/ref-field.component.html
@@ -1,4 +1,4 @@
-<div [id]="path" [ngSwitch]="shouldDisplayWithTemplate" >
+<div [id]="path.join('.')" [ngSwitch]="shouldDisplayWithTemplate" >
   <div *ngSwitchCase="true">
     <button class="btn-preview-template" *ngIf="isLazy" (click)="onPreviewClick($event)">👁</button>
   </div>
--- a/src/ref-field/ref-field.component.html
+++ b/src/ref-field/ref-field.component.html
@@ -1,8 +1,8 @@
 <div [id]="path.join('.')" [ngSwitch]="shouldDisplayWithTemplate" >
   <div *ngSwitchCase="true">
-    <button class="btn-preview-template" *ngIf="isLazy" (click)="onPreviewClick($event)">👁</button>
+    <button class="btn-preview-template" *ngIf="isLazy && !isPreviewButtonHidden" (click)="onPreviewClick($event)">👁</button>
   </div>
   <div *ngSwitchDefault>
-    <a target="_blank" [href]="refValue">{{refValue}}</a>
+    <a target="_blank" [href]="value.get('$ref')">{{value.get('$ref')}}</a>
   </div>
 </div>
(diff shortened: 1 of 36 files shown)


src/ref-field/ref-field.component.html
```diff
@@ -1,8 +1,8 @@
-<div [id]="path" [ngSwitch]="shouldDisplayWithTemplate" >
+<div [id]="path.join('.')" [ngSwitch]="shouldDisplayWithTemplate" >
   <div *ngSwitchCase="true">
-    <button class="btn-preview-template" *ngIf="isLazy" (click)="onPreviewClick($event)">👁</button>
+    <button class="btn-preview-template" *ngIf="isLazy && !isPreviewButtonHidden" [hidden]="isPreviewButtonHidden" (click)="onPreviewClick($event)">👁</button>
```
harunurhan: is `[hidden]` required?

src/table-list-field/table-list-field.component.html
```diff
@@ -13,17 +13,16 @@
...
-					<add-new-element-button *ngIf="schema.items.properties[key].type === 'array'" [key]="key" [schema]="schema.items.properties[key]"
-						[values]="row.get(key) | selfOrEmpty:schema.items.properties[key]" (onNewElementAdd)="onValueChange($event, i, key)"></add-new-element-button>
+					<any-type-field [value]="row.get(key) | selfOrEmpty:schema.items.properties[key]" [schema]="schema.items.properties[key]"
+						[path]="getValuePath(i, key)"></any-type-field>
+					<add-new-element-button *ngIf="schema.items.properties[key].type === 'array'" [path]="getValuePath(i, key)" [schema]="schema.items.properties[key]"></add-new-element-button>
 				</td>
 
 				<!-- UP/DOWN and DELETE buttons for each row -->
-				<td *ngIf="values.size > 0" class="button-holder">
+				<td *ngIf="values.size> 0" class="button-holder">
```
harunurhan: space

harunurhan:
updated! @jmartinm

PR #94, opened 2016-11-30
Complex list item paginate test

harunurhan:


harunurhan:
@jmartinm it is still slow, I think we might have to `debounce` the input event or change it with `blur`.

jmartinm:
> @jmartinm it is still slow, I think we might have to debounce the input event or change it with blur.

It can be changed to on `blur` or the `Enter` key pressed.

harunurhan:
@jmartinm well in that case some "optimizations" in this PR, is a bit redundant and makes the code more complex I think. But I will check how it goes.

example/styles.scss
```diff
@@ -1 +1,3 @@
-/* You can add global styles to this file, and also import other style files */
\ No newline at end of file
+/* You can add global styles to this file, and also import other style files */
+$icon-font-path: '~bootstrap-sass/assets/fonts/bootstrap/';
```
jmartinm: Did this improve something with regard to loading font-awesome icons?

jmartinm:
By using it, some things in the navigation that can be improved:

* Typing on the filter search box, pressing `Enter` should make a search.
* After making a search, the number that appears close to the Up and Down arrows, should reflect the current item.
pagination
* Like in the gif above, ideally you would be able to keep clicking the `Next` button without the window being moved down. One way to do this can be to make the inside of the complex field list scrollable and the outside where the pagination and the search are fixed.

And for the pagination:

* Would be good to see how many references total there are. So something like `Showing 1-10 out of 135`
* And a future improvement can be to allow changing the number of results per page.

example/styles.scss
harunurhan: Unfortunately not, but this was giving `GET` request error while fetching font files before, now there is no error but they aren't displayed :(

src/complex-list-field/complex-list-field.component.html
```diff
@@ -1,28 +1,43 @@
...
+          </span>
+          <input type="search" class="form-control" [(ngModel)]="expression" placeholder="Find" />
+          <span class="input-group-btn">
+            <button type="button" class="btn btn-default" [disabled]="currentFound <= 0" (click)="onFoundNavigate(-1)">❮</button>
+          </span>
+          <span class="input-group-btn">
+            <button type="button" class="btn btn-default" [disabled]="currentFound >= foundIndices.length - 1" (click)="onFoundNavigate(1)">❯</button>
+          </span>
+          <span *ngIf="foundIndices.length > 0" class="input-group-addon">
+            {{foundIndices.length}} out of {{values.size}}
```
harunurhan: @jmartinm  here's the `X out of Y`, where X is number of search results and Y is number of total references.

jmartinm: Yes, so what I was proposing above would be to use `currentFound` instead of `foundIndices.length`

harunurhan:
I also realized that focusing on the next search result now, is very disturbing. But I think making complex list elements scrollable inside might be confusing as well, since we have many independent scrolls on the page. 

Instead, it is a bit bigger task but I was thinking to have something like http://getbootstrap.com/javascript/#scrollspy, that will have the menu for current top field, based on the scroll position. In this way we might get rid of some buttons on the UI too.

jmartinm:
Yes, I remember discussing that in the past and could be an ideal solution. For `table-list` for example it could show the header of the table always so that you can see in which column you are and also add extra columns.

And for complex lists you could see all the controls as you mention above. 👍

harunurhan:
The branch has been changed.

PR #95, opened 2016-12-01
Implement long list navigator.

harunurhan:
* Changes the primitive field to apply the changes to
global store on enter pressed and on input blur.

* Updates tests of primitive-field for blur/press-enter
store update.

* Adds x_editor_long_list_navigator config option.

* Adds navigator when long_list navigator is enabled for
a complext list. Navigator has find functionality and a paginator.

* Removes tsc error for Preview.

TODOS:

* Remove changes in example/*

* Clean the complex list code if possible, might need to  move
some code to a service.

### DIFF
--- a/src/complex-list-field/complex-list-field.component.ts
+++ b/src/complex-list-field/complex-list-field.component.ts
@@ -24,6 +24,7 @@ import {
   Component,
   Input,
   OnChanges,
+  OnInit,
   ChangeDetectionStrategy,
   SimpleChanges
 } from '@angular/core';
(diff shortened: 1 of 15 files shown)


src/complex-list-field/complex-list-field.component.ts
```diff
@@ -42,30 +43,122 @@ import { AppGlobalsService, JsonStoreService } from '../shared/services';
...
+      .concat(index)
+      .join('.');
+    setTimeout(() => this.domUtilService.focusAndSelectFirstInputChildById(itemId));
+  }
+
+  onPageChange(page: number) {
+    this.paginatedIndices = this.getIndicesForPage(page);
+    this.keys = this.getKeysForCurrentPage();
+  }
+
```
harunurhan: Below here could go to a service but, may be we can do it later, when we need to use these elsewhere.

jmartinm:
Just testing and seems to work fine. The only thing that is a bit bothering is that when you press `Enter` in a input there is no feeling that the content was saved. I think it would be better if after pressing `Enter` the focus was changed to the next editable field.

Can it be done already or should we wait for @zzacharo's changes to jump between fields?

harunurhan:
@jmartinm I think we should merge this and create a issue about this to be done after @zzacharo's PR is merged.

PR #98, opened 2016-12-02
fix list related bugs

harunurhan:
* Fixes the bug that occurs  when a new field with `"type": "array"' is
added. The bug is caused by `immutablejs` that inserts always Map in `setIn`
if a key doesn't exist in given keyPath.

* Fixes the bug with primitive-list, it displays values correctly.


jmartinm:
Seems to work as expected.

Is it possible to add a test for that new code?

PR #100, opened 2016-12-05
shortcuts: Creation of keyboard shortcuts

zzacharo:


zzacharo:
@jmartinm Must rebase and push again. @harunurhan commit was picked just for continue with implementation. My commit is just for initial review.

zzacharo:
@jmartinm @harunurhan https://travis-ci.org/inveniosoftware-contrib/ng2-json-editor/jobs/181606027#L2336 any idea why it still cannot find these modules? Locally it works but the tests are failing too when running `npm test`.

src/shared/services/tab-index.service.ts
```diff
@@ -0,0 +1,90 @@
...
+ * as an Intergovernmental Organization or submit itself to any jurisdiction.
+ */
+
+import { Injectable } from '@angular/core';
+import { ReplaySubject } from 'rxjs';
+
+
+@Injectable()
+export class TabIndexService {
+  private elemTabIndexList = Array<string>();
```
harunurhan: Not important but can you replace `elem` with `field` because I think it makes more sense, `elem` is used for dom element variables.

src/shared/services/shortcuts.service.ts
```diff
@@ -0,0 +1,112 @@
...
+ * as an Intergovernmental Organization or submit itself to any jurisdiction.
+ */
+
+
+import { Injectable } from '@angular/core';
+import { TabIndexService } from './tab-index.service';
+import { ShortcutsUtilService } from './shortcuts-util.service';
+
+@Injectable()
+export class ShortcutsService {
```
harunurhan: Also, it kind of bothers me that we have to repeat
```
event.preventDefault();
let eventTarget = event.target as HTMLInputElement;
....actionCall...
return false;
```

maybe we could have a generator function that takes the action name then returns a new function that calls
the action with given name. To do this you need to change the other service too.

src/shared/directives/hotkeys.directive.ts
```diff
@@ -0,0 +1,53 @@
...
+  Directive,
+  OnInit, Input, OnDestroy
+} from '@angular/core';
+
+import 'mousetrap';
+
+@Directive({
+  selector: '[hotkeys]'
+})
+export class HotkeysDirective implements OnInit, OnDestroy {
```
harunurhan: let's remove it to `ShortcutsDirective`, since this our own implementation.

src/shared/directives/hotkeys.directive.ts
```diff
@@ -0,0 +1,53 @@
...
+  @Input('hotkeys') hotkeys: {string: (event: KeyboardEvent) => boolean};
```
harunurhan: I think param for `@Input` is unnecessary when you have the variable name is the same.

src/shared/services/dom-util.service.ts
```diff
@@ -32,6 +32,41 @@ export class DomUtilService {
       if (firstInput) {
         firstInput.focus();
         firstInput.select();
+      }else {
```
harunurhan: space after `}`, btw this should fail the build because of our `tslint` rules.

src/shared/services/dom-util.service.ts
```diff
@@ -32,6 +32,41 @@ export class DomUtilService {
...
+          firstInput.classList.add('open');
+          firstInput = el.querySelector('div.btn-group input') as HTMLInputElement;
+          firstInput.focus();
+          firstInput.select();
+        }
+      }
+    }
+  }
+
+  focusInputByTabindex(index: number) {
```
harunurhan: can't you get the `id` with tab index since we have all `id`s sorted in an array, then select by id. This could be be faster than this.

src/shared/services/json-schema.service.ts
```diff
@@ -0,0 +1,60 @@
...
+ */
+
+import { Injectable } from '@angular/core';
+
+@Injectable()
+export class JsonSchemaService {
+
+  private schema: Object;
+
+  constructor() { }
```
harunurhan: no need for an empty constructor.

src/shared/services/shortcuts-util.service.ts
```diff
@@ -0,0 +1,180 @@
...
+  }
+
+  /**
+   * @param {Array<any>} path - Path of the array parent that the element is about to be inserted
+   * @param {Object} schema - Schema of the element that is about to be inserted
+   */
+  addNewElementInArray(path: Array<any>, schema: Object) {
+    let itemSchema = schema['items'];
+    let emptyValue = this.emptyValueService.generateEmptyValue(itemSchema);
+    let values = this.jsonStoreService.getIn(path) || List.of([]);
```
harunurhan: `|| List()` is enough.

src/shared/services/shortcuts-util.service.ts
```diff
@@ -0,0 +1,180 @@
...
+    console.log(originalPath);
+    console.log(elemIndex);
+    emptyValue = value;
+    emptyValue = emptyValue.set(originalPath[originalPath.length - 1], '');
+    this.jsonStoreService.setIn(arrayParentPath, valuesList.insert(elemIndex + 1, emptyValue));
+    arrayParentPath.push(elemIndex + 1);
+    setTimeout(() => {
+      this.tabIndexService.insertElemTabIndex(arrayParentPath.join('.').concat('.' + originalPath[originalPath.length - 1]));
+      this.domUtilService.flashElementById(arrayParentPath.join('.').concat('.' + originalPath[originalPath.length - 1]));
+      this.domUtilService.focusAndSelectFirstInputChildById(arrayParentPath.join('.').concat('.' + originalPath[originalPath.length - 1]));
```
harunurhan: call join('.') once (possibly outside of timeout) and then reuse it instead.

src/table-list-field/table-list-field.component.ts
```diff
@@ -63,6 +63,8 @@ export class TableListFieldComponent extends AbstractListFieldComponent implemen
 
   onFieldAdd(field: string) {
     this.keys.push(field);
+    setTimeout(() => {
```
zzacharo: Because it is needed for ensuring that the new components that are about to be created where first added to the `elemTabIndexList`. But logically it doesn't make sense to move it to the `sortAndSynchronizeTabIndexes`.

src/shared/services/shortcuts-util.service.ts
```diff
@@ -0,0 +1,180 @@
...
+import { EmptyValueService } from './empty-value.service';
+import { JsonStoreService } from './json-store.service';
+import { JsonSchemaService } from './json-schema.service';
+import { DomUtilService } from './dom-util.service';
+import { TabIndexService } from './tab-index.service';
+import { PathUtilService } from './path-util.service';
+import { List } from 'immutable';
+
+@Injectable()
+export class ShortcutsUtilService {
```
zzacharo: You mean the methods that are used internallly by `ShortcutUtilService`. Beacuse the shortcut methods must be public in order to be used by other components.

harunurhan:
@zzacharo For tests to pass, you need to the `TabIndexService` to  `primitive-field.component.spec.ts` like the other services.

src/shared/services/shortcuts-util.service.ts
harunurhan: Sorry I meant the ones that doesn't have the `Shortcut` prefix, so yep internally used methods.

src/shared/services/shortcut-action.service.ts
```diff
@@ -0,0 +1,276 @@
...
+    });
+  }
+
+  private updateTabIndexesAndFocusElementInPath(path: string) {
+    this.tabIndexService.insertElemTabIndex(path);
+    this.domUtilService.flashElementById(path);
+    this.domUtilService.focusAndSelectFirstInputChildById(path);
+  }
+
+  generateShortcutAction(action: string) {
```
zzacharo: @harunurhan this generator function will return every callback function based on given action. Is it ok?

harunurhan: unfortunately not, because those 3 lines are still repeated. so I was asking for something like: 

```typescript
class ShortcutActionService {

  private deleteAction(path: string) {

  }
  
  private addAction(path: string) {

  }

  constructor(private tabIndexService: TabIndexService) { }

  public getActionByName(actionName: string) {
    return (event: KeyboardEvent) => {
      event.preventDefault();
      let targetInput = event.target as HTMLInputElement;
      this[actionName](this.tabIndexService.getTabIndexFromTabIndex(targetInput.tabIndex));
      return false;
    };
  }

}

class TabIndexService {
  public getTabIndexFromTabIndex(tabIndex: number) {
    return 'path for tabindex';
  }
}

class ShortcutService {
  constructor(private shortcutActionService: ShortcutActionService) { }

  shortcuts = {
    'delete': {
      key: 'mod+del',
      action: this.shortcutActionService.getActionByName('deleteAction')
    },
    'add': {
      key: 'mod+a',
      action: this.shortcutActionService.getActionByName('addAction')
    }
  }
}
```

src/table-list-field/table-list-field.component.ts
harunurhan: OK, but since this function doesn't access the DOM, why do we wait for component to be created. Can't we just set its `path` and `tabIndex` to the map and then the new component can access it during initialisation.

zzacharo:
@harunurhan travis build was successfull. So, tests are remaining to be covered. You can review my PR again.

src/table-list-field/table-list-field.component.ts
zzacharo: No, actually this ensures that the `sortAndSynchronize` method in shortcutActionService will be called after all new elements are created and pushed in the `elemTabIndexList`. In this way, i can sort only one time. Otherwise i must trigger sorting every time a new element is added. That's why i insert only a new empty element in parent array( for example a new item in authors such as authors.1) and all children of this element are created from angular bindings.

README.md
```diff
@@ -42,6 +42,7 @@ Please have a look at [example app](./example/app)
 | ------------- |-------------|-------------|
 | *schemaOptions*      | Object | Enriches JsonSchema with custom properties that `ng2-json-editor` understands. See [possible values](#schemaOptions).|
 | *previews*      | Array | Configures previews for the document. See [configuration syntax](#previews).|
+| *shortcuts*     | Object | Configures shortcuts for the document. See [shortcuts_syntax](#shortcuts).|
```
harunurhan: shortcuts syntax.

src/json-editor.component.html
```diff
@@ -13,7 +13,7 @@
       </tr>
       <tr>
         <td>
-          <any-type-field [value]="_record.get(key)" [schema]="schema.properties[key]" [path]="getPathForField(key)"></any-type-field>
+          <any-type-field [value]="_record.get(key) | selfOrEmpty:schema.properties[key]" [schema]="schema.properties[key]" [path]="getPathForField(key)"></any-type-field>
```
harunurhan: is it relevant?

src/json-editor.component.ts
```diff
@@ -97,13 +102,14 @@ export class JsonEditorComponent extends AbstractTrackerComponent implements OnI
...
       });
+    this.jsonSchemaService.setSchema(this.schema);
   }
 
   /**
    * Extracts previews from config, and populates url if necessary
    * by using `getUrl` or `urlPath` configs and the record.
    */
-  extractPreviews(): Array<any> {
+  public extractPreviews(): Array<any> {
```
harunurhan: I think this should be even `private`

src/primitive-field/primitive-field.component.html
```diff
@@ -2,7 +2,7 @@
   <div class="editable-field-container" [ngClass]="errorNgClass" [tooltipHtml]="errors | errorsToMessagesHtml" tooltipTrigger="mouseenter"
     [tooltipEnable]="isErrorTooltipEnabled">
     <div *ngSwitchCase="'string'">
-      <textarea rows="1" textareaAutofit [(ngModel)]="value" (blur)="commitValueChange()" (keypress)="onKeypress($event)" placeholder="{{schema.title}}"></textarea>
+      <textarea rows="1" [tabindex]="tabIndex" textareaAutofit [(ngModel)]="value" (blur)="commitValueChange()" (keypress)="onKeypress($event)" placeholder="{{schema.title}}"></textarea>
```
harunurhan: I think, all input field here should have `tabIndex`, at least the `searchable-dropdown` and `autocomplete-input.

@jmartinm

src/shared/services/json-schema.service.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+          }
+        },
+        'type': 'object'
+      },
+      'type': 'array',
+      'uniqueItems': true
+    };
+    let path = ['abstracts', 0];
+    service.setSchema(schema);
+    expect(JSON.stringify(service.getSchemaFromPath(path))).toBe(JSON.stringify(expectedSchema));
```
harunurhan: use `toEqual` without `JSON.stringfy` to compare `object`s

src/shared/services/path-util.service.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+    let value = 'value';
+    let json = fromJS({
+      aMap: {}
+    });
+    jsonStoreService.setJson(json);
+    jsonStoreService.setIn(path, value);
+    let expectedRootParent = ['aMap', 'aList'];
+    let expectedNearestParent = ['aMap', 'aList', 0, 'innerList'];
+    expect(JSON.stringify(pathUtilService.getNearestOrRootArrayParentInPath(path, true))).toBe(JSON.stringify(expectedRootParent));
+    expect(JSON.stringify(pathUtilService.getNearestOrRootArrayParentInPath(path, false))).toBe(JSON.stringify(expectedNearestParent));
```
harunurhan: ditto

src/shared/services/shortcut-action.service.ts
```diff
@@ -0,0 +1,200 @@
...
+  }
+
+  moveDownAction(path: Array<any>): void {
+    this.moveAction(path, 1);
+  }
+
+  private moveAction(path: Array<any>, direction: number): void {
+    let index = this.pathUtilService.getElementIndexInReversePath(path);
+    path[path.length - 2] = this.moveElement(index, direction, this.pathUtilService.getNearestOrRootArrayParentInPath(path, false));
+    let _path = path.join('.');
```
harunurhan: not important but `pathString` is better name.

src/shared/services/shortcut-action.service.ts
```diff
@@ -0,0 +1,200 @@
...
+
+  navigateUpAction(path: Array<any>): void {
+    this.navigateUpDownAction(path, -1);
+  }
+
+  navigateDownAction(path: Array<any>): void {
+    this.navigateUpDownAction(path, 1);
+  }
+
+  private navigateUpDownAction(path: Array<any>, direction: number) {
```
harunurhan: shouldn't have `Action` postfix

src/shared/services/shortcut-action.service.ts
```diff
@@ -0,0 +1,200 @@
...
+
+  navigateLeftAction(path: Array<any>): void {
+    this.navigateRightLeftAction(path, -1);
+  }
+
+  navigateRightAction(path: Array<any>): void {
+    this.navigateRightLeftAction(path, 1);
+  }
+
+  private navigateRightLeftAction(path: Array<any>, direction: number) {
```
harunurhan: shouldn't have `Action` postfix

src/shared/services/shortcut-action.service.ts
```diff
@@ -0,0 +1,200 @@
...
+    this.jsonStoreService.setIn(arrayParentPath, valuesList.insert(elemIndex + 1, newValue));
+    arrayParentPath.push(elemIndex + 1);
+    setTimeout(() => {
+      this.updateTabIndexesAndFocusElementInPath(arrayParentPath.join('.').concat('.' + originalPath[originalPath.length - 1]));
+    });
+  }
+
+  private copyCurrentSchemaItemBelow(originalPath: Array<any>, arrayParentPath: Array<any>) {
+    let elemIndex = this.pathUtilService.getElementIndexInPath(originalPath);
+    let valuesList = this.jsonStoreService.getIn(arrayParentPath) || List.of([]);
```
harunurhan: `List()` instead of `List.of([])`

src/shared/services/shortcut-action.service.ts
```diff
@@ -0,0 +1,200 @@
...
+  }
+
+  copyItemAction(path: Array<any>) {
+    let _path = this.pathUtilService.getNearestOrRootArrayParentInPath(path, true);
+    this.copyCurrentSchemaItemBelow(path, _path);
+  }
+
+  private copyCurrentRowBelow(originalPath: Array<any>, arrayParentPath: Array<any>) {
+    let elemIndex = this.pathUtilService.getElementIndexInReversePath(originalPath);
+    let valuesList = this.jsonStoreService.getIn(arrayParentPath) || List.of([]);
```
harunurhan: List() instead of List.of([])

src/shared/services/shortcut-action.service.ts
```diff
@@ -0,0 +1,200 @@
...
+    let newValue = valuesList.get(elemIndex);
+    newValue = newValue.set(originalPath[originalPath.length - 1], '');
```
harunurhan: why are you setting empty string here?

src/shared/services/shortcut.service.ts
```diff
@@ -0,0 +1,96 @@
...
+
+
+import { Injectable } from '@angular/core';
+import { TabIndexService } from './tab-index.service';
+import { ShortcutActionService } from './shortcut-action.service';
+
+@Injectable()
+export class ShortcutService {
+
+  shortcuts = {
```
harunurhan: I think shortcut keys  should be same as action name but without `Action` prefix, like: `addParent`, `moveUp` etc. (after this, README update is necessary)

And you don't have to put as a string you can do
```
{
  addParent: {
    key: ...
    action: ...
  }...
```

src/shared/services/shortcut-action.service.ts
```diff
@@ -0,0 +1,200 @@
...
+import { EmptyValueService } from './empty-value.service';
+import { JsonStoreService } from './json-store.service';
+import { JsonSchemaService } from './json-schema.service';
+import { DomUtilService } from './dom-util.service';
+import { TabIndexService } from './tab-index.service';
+import { PathUtilService } from './path-util.service';
+import { List } from 'immutable';
+
+@Injectable()
+export class ShortcutActionService {
```
harunurhan: every action should be private for now at least.

src/shared/services/shortcut-action.service.ts
```diff
@@ -0,0 +1,200 @@
...
+
+  addRootAction(path: Array<any>): void {
+    this.addAction(path, true);
+  }
+
+  addParentAction(path: Array<any>): void {
+    this.addAction(path, false);
+  }
+
+  private addAction(path: Array<any>, root: boolean): void {
```
harunurhan: shouldn't have `Action` postfix.

src/shared/services/shortcut-action.service.ts
```diff
@@ -0,0 +1,200 @@
...
+
+  moveUpAction(path: Array<any>): void {
+    this.moveAction(path, -1);
+  }
+
+  moveDownAction(path: Array<any>): void {
+    this.moveAction(path, 1);
+  }
+
+  private moveAction(path: Array<any>, direction: number): void {
```
harunurhan: shouldn't have `Action` prefix

src/shared/services/shortcut-action.service.ts
```diff
@@ -0,0 +1,200 @@
...
+  }
+
+  /**
+   * @param {Array<any>} path - Path of the array parent that the element is about to be inserted
+   * @param {Object} schema - Schema of the element that is about to be inserted
+   */
+  private addNewElementInArray(path: Array<any>, schema: Object) {
+    let itemSchema = schema['items'];
+    let emptyValue = this.emptyValueService.generateEmptyValue(itemSchema);
+    let values = this.jsonStoreService.getIn(path) || List.of([]);
```
harunurhan: `List()` instead of `List.of([])`

Also is it possible to reuse `AbstractListField`'s add new element functionality here.

src/primitive-field/primitive-field.component.html
jmartinm: Yes, the only with doubt is the boolean. For the moment we can add it all of them and we can see if later on we remove the boolean one.

src/shared/services/shortcut-action.service.ts
```diff
@@ -0,0 +1,200 @@
...
+    this.domUtilService.focusAndSelectFirstInputChildById(path);
+  }
+
+  generateShortcutAction(action: string) {
+
+    return (event: KeyboardEvent) => {
+      event.preventDefault();
+      let eventTarget = event.target as HTMLInputElement;
+      this[action](this.tabIndexService.getPathFromTabIndex(eventTarget.tabIndex));
+      return false;
```
jmartinm: Do you need both `event.preventDefault();` and `return false` ?

src/shared/services/shortcut.service.ts
```diff
@@ -0,0 +1,96 @@
...
+  constructor(public tabIndexService: TabIndexService,
+              public shortcutActionService: ShortcutActionService) { }
+
+  getShortcuts(config: EditorConfig) {
+    return this.enrichShortcutsWithConfig(config);
+  }
+
+  enrichShortcutsWithConfig(config: EditorConfig): Object {
+    let shotcutsFromConfig = config['shortcuts'];
+    if (shotcutsFromConfig) {
```
jmartinm: typo `shotcutsFromConfig` -> `shortcutsFromConfig`

src/shared/services/path-util.service.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+  });
+
+  it('should return the right index in direct or reverse given path', () => {
+    let path = ['aMap', 'aList', 0, 'innerList', 0, 'innerProperty'];
+    let expectedDirectIndex = 0;
+    let expectedReverseIndex = 0;
```
jmartinm: Would be better to use different index values to make sure that the function is really doing the right thing.

src/shared/services/shortcut-action.service.ts
```diff
@@ -0,0 +1,200 @@
...
+      this.domUtilService.flashElementById(_path);
+      this.domUtilService.focusAndSelectFirstInputChildById(_path);
+    });
+  }
+
+  /**
+   * @param {number} index - Index of the element that is moved
+   * @param {number} direction - Movement direction. -1 for UP, +1 for DOWN
+   * @returns {number} - Returns the new index of the moved element
+   */
```
jmartinm: These docstrings are very helpful 👍

src/shared/services/shortcut-action.service.ts
harunurhan: I think first one is disable the default action, and the return is for bubbling, since they might have pressed the shortcut for longer time.

src/json-editor.component.html
zzacharo: no it was a leftover from when i was trying to implement top level fields addition with the button below tree menu.

src/shared/services/shortcut-action.service.ts
zzacharo: Gets an unused method error by tslint during building when i set the actions to private.

src/shared/services/shortcut-action.service.ts
zzacharo: @harunurhan like calling directly `AbstractListField` 's method?

zzacharo:
@harunurhan @jmartinm Pushed again with review changes.

src/shared/services/dom-util.service.ts
```diff
@@ -32,6 +32,35 @@ export class DomUtilService {
...
+          firstInput = el.querySelector('div.btn-group input') as HTMLInputElement;
+          firstInput.focus();
+          firstInput.select();
+        }
+      }
+    }
+  }
+
+  // TO-DO: find the next editable element in row.
+  focusRightOrLeftCell(index: number, direction: number) {
```
harunurhan: `focusRightOrLeftParentCellById(id: string, direction: number)`

src/shared/services/dom-util.service.ts
```diff
@@ -32,6 +32,35 @@ export class DomUtilService {
...
+    let firstInput = document.querySelector('[tabIndex=\'' + index + '\']') as HTMLElement;
```
harunurhan: `let element = document.getElementById(id)`

src/shared/services/dom-util.service.ts
```diff
@@ -32,6 +32,35 @@ export class DomUtilService {
...
+    if (firstInput) {
+      while (firstInput.nodeName !== 'TD') {
+        firstInput = firstInput.parentElement;
+      }
+      let nextSibling = direction > 0 ? firstInput.nextElementSibling : firstInput.previousElementSibling;
+      while (nextSibling && nextSibling.nodeName === 'TD') {
+        let textElem = nextSibling.querySelector('div.editable-field-container input,textarea') as HTMLInputElement;
```
harunurhan: `inputElement` instead of `textElem`

src/shared/services/dom-util.service.ts
```diff
@@ -32,6 +32,35 @@ export class DomUtilService {
...
+      }
+    }
+  }
+
+  // TO-DO: find the next editable element in row.
+  focusRightOrLeftCell(index: number, direction: number) {
+    let firstInput = document.querySelector('[tabIndex=\'' + index + '\']') as HTMLElement;
+    if (firstInput) {
+      while (firstInput.nodeName !== 'TD') {
+        firstInput = firstInput.parentElement;
```
harunurhan: define a local variable outside of `while` like `parentCell` and assign`.parentElement` to it here. In that case you avoid confusion that is caused by `firstInput` variable being a`td`.

src/shared/services/shortcut-action.service.ts
```diff
@@ -0,0 +1,232 @@
...
+
+  /**
+   * @param {Array<any>} path - Path of the element that is focused
+   * @param {number} direction - Navigation direction. -1 for LEFT, +1 for RIGHT
+   */
+  private navigateRightLeft(path: Array<any>, direction: number) {
+    let tabIndex = this.tabIndexService.getElemTabIndex(path.join('.'));
+    if (tabIndex >= 0) {
+      this.domUtilService.focusRightOrLeftCell(tabIndex, direction);
+    }
```
harunurhan: first see the comment for `domUtilService.focusRightOrLeftCell`.

After that you can reduce this, to

`this.domUtilService.focusRightOrLeftParentCellById(path.join('.'), direction);`

src/shared/services/shortcut-action.service.ts
```diff
@@ -0,0 +1,232 @@
...
+    let temp = values.get(index);
+    values = values
+      .set(index, values.get(newIndex))
+      .set(newIndex, temp);
+    this.jsonStoreService.setIn(path, values);
+    return newIndex;
+  }
+
+  deleteAction(path: Array<any>) {
+    if (window.confirm('Are you sure you want to delete this item?')) {
```
jmartinm: We can remove it since the button to Delete will now be hidden under a Dropdown and then there is undo, so an accident can be recovered.

src/shared/services/dom-util.service.ts
zzacharo:
```
let elementParentCell = el.parentElement;
      while (elementParentCell.nodeName !== 'TD') {
        elementParentCell = elementParentCell.parentElement;
      }
```
You mean something like this???

jmartinm:
At the moment this PR has conflicts with master branch, can you do a rebase so that it can be tested locally?

zzacharo:
@jmartinm @harunurhan Rebased. I made the changes. Please review again.

jmartinm:
For some reason when using tab in between fields, in some cases the value gets selected and in others it is not:

select-tab

zzacharo:
@jmartinm strange i will check it now.

jmartinm:
In `accelerator_experiments` if you are in the `experiment` field and do `Ctrl+left` it jumps to the boolean field but then `Ctrl+right` does not work to go back to the previous field.

I think the best behaviour is to allow it to jump to the boolean field but then from there that shortcut should still work.

src/shared/services/dom-util.service.ts
zzacharo: @harunurhan after changed this it gets more complicated because using tabindex i could directly identify the input or textarea field that is actually editable. Otherwise with the path i get the div parent and it needs to check more stuff in DOM to check if the next/previous element is tabable. This change actually leads to @jmartinm comment about the ctrl + right/ctrl + left bug. Should i use the previous implementation?

zzacharo:
@jmartinm pushed again. Changed navigation to only tabable elements. Do we want to check for the checkboxes? Because now they are not tabable.

src/shared/services/path-util.service.ts
```diff
@@ -0,0 +1,70 @@
...
+@Injectable()
+export class PathUtilService {
+
+  /**
+   *
+   * @param {Array<any>} path - Element's path
+   * @param {boolean} root - Find nearest or root parent array. True for root and false for nearest array parent
+   * @returns {Array<any>} - Returns the path to the array parent
+   */
+  getNearestOrRootArrayParentInPath(path: Array<any>, root: boolean): Array<any> {
```
jmartinm: There are many cases in this PR where you use a flag in the method to separate the behaviour. You can read here <http://martinfowler.com/bliki/FlagArgument.html> an argument of why `flag` arguments are not so good.

While reading the calls to these methods, like `getNearestOrRootArrayParentInPath(path, true)` it's difficult to know what's going to happen.

So I would suggest refactoring them if possible into two different methods:

`getNearestArrayParentInPath` and `getRootArrayInPath` that just accept a `path` parameter. The common code in those functions you can extract into a helper function and leave inside of them just the differing code.

Apply this to all the cases where you have this pattern of using one function with a flag that changes the behaviour inside and it will improve a lot the readability.

src/shared/services/shortcut-action.service.ts
```diff
@@ -0,0 +1,219 @@
...
+
+  constructor(public emptyValueService: EmptyValueService,
+              public domUtilService: DomUtilService,
+              public jsonStoreService: JsonStoreService,
+              public jsonSchemaService: JsonSchemaService,
+              public pathUtilService: PathUtilService,
+              public tabIndexService: TabIndexService) { }
+
+  addToRootAction(path: Array<any>): void {
+    this.add(path, true);
```
jmartinm: See comment above about flag argument.

jmartinm:
When doing the shortcut that adds a top level `reference` the dropdown to add field opens but its quite broken. Typing filters values but cannot navigate with the keyboard down, also clicking outside does not close the dropdown.

add-new

src/shared/services/shortcut.service.ts
```diff
@@ -0,0 +1,95 @@
...
+    moveUp: {
+      key: 'mod+shift+up',
+      action : this.shortcutActionService.generateShortcutAction('moveUpAction')
+    },
+    moveDown: {
+      key: 'mod+shift+down',
+      action: this.shortcutActionService.generateShortcutAction('moveDownAction')
+    },
+    delete: {
+      key: 'mod+del',
```
jmartinm: This combination does not work on Mac it seems.

jmartinm:
Pressing `alt+c` in references -> misc gives an error in the JS console.

example/app/app.config.ts
```diff
@@ -24,6 +24,14 @@ export const EXAMPLE_CONFIG: AppConfig = {
         return `http://arxiv.org/abs/${value}`;
       }
     }
+  },
+  shortcuts: {
+    'navigateLeft': {
+      key: 'ctrl+left'
```
zzacharo: @jmartinm just an example configuration.

src/shared/services/dom-util.service.ts
```diff
@@ -32,6 +32,35 @@ export class DomUtilService {
       if (firstInput) {
         firstInput.focus();
         firstInput.select();
+      } else {
+        firstInput = el.querySelector('div.btn-group') as HTMLInputElement;
+        if (firstInput) {
+          // .open is used for opening the dropdownmenu
+          firstInput.classList.add('open');
+          firstInput = el.querySelector('div.btn-group input') as HTMLInputElement;
+          firstInput.focus();
```
zzacharo: @jmartinm i don't think this is the same logic as before. If first input is undefined everywhere it will attempt to focus and select the undefined element.

jmartinm: Amended the example, the main thing is to aggregate in the first part of the function the looking for the input, and at the end, if you found it, do the lines:

```javascript
firstInput.focus();
firstInput.select();
```

src/shared/services/path-util.service.ts
```diff
@@ -0,0 +1,70 @@
...
+    let _path = [];
+    let pathLength = path.length;
+    for (let index = 0; index < pathLength; index++) {
+      _path = root ? path.slice(0, index + 1) : path.slice(0, path.length - index);
+      if (typeof _path[_path.length - 1] === 'number') {
+        // transform ['arrayParent',0] => ['arrayPrent']
```
jmartinm: Ah ok, it's fine I think. There is just a typo in the second part `['arrayPrent']` -> `['arrayParent']`

src/shared/services/path-util.service.ts
zzacharo: @jmartinm this flag was used to squash the code. In the previous review i had some actions splitted but if turn back all the flags it will expand the code. Because yesterday i was trying to squash the code. If instead i use an upper level function that uses internally the flags? Like
```
addToRootAction(path: Array<any>): void {
    this.add(path, true);
```
The functions with flags are private so with this way they are hidden from other components.

jmartinm: In the case you mention it is not that bad since the public facing function is clear `addToRootAction`.

But in other cases, it is not very readable, like the public facing ones - https://github.com/inveniosoftware-contrib/ng2-json-editor/pull/100/files#diff-a896742e125e064d80fd8e9c3aa951d5R34 so I would amend those to favor readability of the code. Otherwise one has to be checking the function signature all the time to see what the call is doing.

src/shared/services/shortcut.service.ts
zzacharo: Do you want to change the default one?

zzacharo:
@jmartinm I pushed again code. You can check if the adding to references and the dropdown is functional now.

jmartinm:
> Pressing alt+c in references -> misc gives an error in the JS console.

This still gives the error:

```javascript
error_handler.js:53 TypeError: newValue.set is not a function
    at ShortcutActionService.copyRowOrSchemaBelow (shortcut-action.service.ts:194)
    at ShortcutActionService.copyAction (shortcut-action.service.ts:170)
    at shortcut-action.service.ts:214
    at _fireCallback (mousetrap.js:610)
    at Mousetrap.self._handleKey (mousetrap.js:671)
    at Mousetrap.handleKey (mousetrap.js:988)
    at HTMLDivElement._handleKeyEvent (mousetrap.js:731)
    at ZoneDelegate.invokeTask (zone.js:265)
    at Object.onInvokeTask (ng_zone.js:227)
    at ZoneDelegate.invokeTask (zone.js:264)
```

jmartinm:
There is some strange behaviour with the delete shortcut and the affiliations field:

del-shortcut

After deleting, it reappears on its own.

jmartinm:
> You can check if the adding to references and the dropdown is functional now.

Dropdown seems to work fine 👍

zzacharo:
@jmartinm the bug is at  the `misc` field in `references` because in schema.json there is this 
```
"misc": {
                "items": {
                  "type": "string"
                }, 
                "type": "array"
              }, 
```
I thought that all arrays have items->properties fields.

zzacharo:
@jmartinm fixed the delete issue with focus out before deleting. You can test it again.

zzacharo:
@jmartinm Updated changes. You can check again.

jmartinm:
Pending unit and protractor tests (see https://github.com/inveniosoftware-contrib/ng2-json-editor/issues/128)

e2e/shortcut-action/shortcut-action.e2e-spec.ts
```diff
@@ -0,0 +1,81 @@
...
+
+import { ShortcutActionPage } from './shortcut-action.po';
+
+describe('Shortcut-action App', function() {
+  let page: ShortcutActionPage;
+  let hotkeys;
+
+  beforeEach(() => {
+    page = new ShortcutActionPage();
+    hotkeys = require('protractor-hotkeys');
```
harunurhan: can't we use `import` like all other modules.

README.md
```diff
@@ -247,6 +248,52 @@ Note that:
 - Either `url`, `getUrl`, or `urlPath` must be present.
 - `html` type uses `iframe` and can display `.pdf` if supported by browser.
 
+### <a name="shortcuts"></a>Shortcuts
+
+Configuration for the shortcuts in the document based on focused fields.
+
+```
+{
+ shortcut-name: {
```
zzacharo: @harunurhan I tried to use the helper function in the `environment.ts` file but from protractor you cannot access the window object so it gives an error. By executing a script with `browser.executescript()` function you can inject a script in the current open browser of the test but in that case i cannot import my file. So, I just use the this approach to obtain the information before every test scenario from the window object. You can check it in my latest commit here:
https://github.com/inveniosoftware-contrib/ng2-json-editor/pull/100/files#diff-5a31329b1c08e4b38445f4db7d5b13aeR18. Any thoughts?

zzacharo: @harunurhan For the bug this is the problem as described in this post:
http://stackoverflow.com/questions/6761169
I am trying to set explicit focus on checkbox on click.

e2e/shortcut-action/shortcut-action.e2e-spec.ts
```diff
@@ -0,0 +1,86 @@
...
+    expect(firstColAfterCopyInputElemValue).toEqual('');
+    expect(secondColAfterCopyInputElemValue).toEqual(secondColInputElemValue);
+  });
+
+  it('should copy new row under references table using `ctrl+shift+c` shortcut. ' +
+    'It must copy the exact value of the root element eg Copy the whole author under the focused one.', () => {
+    let inputElemValue = page.getInputElementById('authors.0.affiliations.0.value').getAttribute('value');
+    hotkeys.trigger(`${mod}+shift+c`, { targetElement: inputElemValue });
+    let afterCopyInputElemValue = page.getInputElementById('authors.1.affiliations.0.value').getAttribute('value');
+    expect(inputElemValue).toEqual(afterCopyInputElemValue);
```
jmartinm: You can also count that the number of `input` and `textarea` equals the element copied.

e2e/shortcut-action/shortcut-action.po.ts
```diff
@@ -0,0 +1,15 @@
...
+export class ShortcutActionPage extends Ng2JsonEditorPage {
+
+  getNumberOfTextareaElementsInInnerTable(id: string) {
+    let tableElem = element(by.id(id));
+    return tableElem.all(by.css('tr[id*=\'' + id + '\']')).count();
+  }
+
+  getNumberOfTextareaElementsInRootTable(id: string) {
+    let tableElem = element(by.id(id));
+    return tableElem.all(by.css('div+div')).count();
```
jmartinm: Instead of `div+div` you can count the number of `table` elements with id like the parameter.

e2e/shortcut-action/shortcut-action.po.ts
```diff
@@ -0,0 +1,15 @@
...
+import { Ng2JsonEditorPage } from '../app.po';
+
+export class ShortcutActionPage extends Ng2JsonEditorPage {
+
+  getNumberOfTextareaElementsInInnerTable(id: string) {
+    let tableElem = element(by.id(id));
+    return tableElem.all(by.css('tr[id*=\'' + id + '\']')).count();
+  }
+
+  getNumberOfTextareaElementsInRootTable(id: string) {
```
jmartinm: Better name can be `getNumberOfChildTablesbyId`

e2e/shortcut-action/shortcut-action.po.ts
```diff
@@ -0,0 +1,15 @@
+import { element, by } from 'protractor/globals';
+import { Ng2JsonEditorPage } from '../app.po';
+
+export class ShortcutActionPage extends Ng2JsonEditorPage {
+
+  getNumberOfTextareaElementsInInnerTable(id: string) {
```
jmartinm: `getNumberOfChildRowsbyId`

e2e/app.po.ts
```diff
@@ -1,11 +1,12 @@
...
 
 export class Ng2JsonEditorPage {
+
   navigateTo() {
     return browser.get('/');
   }
 
-  getParagraphText() {
-    return element(by.css('app-root h1')).getText();
+  getInputElementById(id: string) {
```
jmartinm: Either `getTextareaChildrenById` or separate into 2 different methods (one to get element by id and the other one to get textareas inside of that element)

jmartinm:
This PR is also missing integration with Travis to run the tests.

e2e/shortcut-action/shortcut-action.e2e-spec.ts
```diff
@@ -0,0 +1,119 @@
...
+    hotkeys.trigger('alt+c', { targetElement: firstColInputElemValue });
+    let firstColAfterCopyInputElemValue = page
+      .getChildOfElementByCss(page.getElementById('keywords.1.classification_scheme'), 'textarea')
+      .getAttribute('value');
+    let secondColAfterCopyInputElemValue = page.getChildOfElementByCss(page.getElementById('keywords.1.keyword'), 'textarea')
+      .getAttribute('value');
+    expect(firstColAfterCopyInputElemValue).toEqual('');
+    expect(secondColAfterCopyInputElemValue).toEqual(secondColInputElemValue);
+  });
+  it(`should copy new row under references table using 'mod+shift+c' shortcut.
```
harunurhan: this fails because, `command+shift+c` is a valid Chrome shortcut on `macOS`. So we should change it.

zzacharo:
@harunurhan @jmartinm In this commit 0e27f438d28cd949280d7744d4f8c5ce8545c9a4 I changed just the keyboard shortcut to `mod+alt+r`. Check if it's fine in Mac and I will continue to work on making travis work with mac env.

harunurhan:
@zzacharo can you rebase this and test it to make not tested parts don't break after a lot of updates?

zzacharo:
@harunurhan  i will rebase and push one commit with the successfull build without mac suppport on travis for now.

zzacharo:
@jmartinm @harunurhan I will work a little bit on this today and if still not working I will refactor it and push again. I ended up that the problem is something with Chrome and Karma on MacOS on travis. Checking this issue:
https://github.com/karma-runner/karma/issues/635 .

zzacharo:
@harunurhan @jmartinm https://travis-ci.org/inveniosoftware-contrib/ng2-json-editor/jobs/194531253#L150 . Node 6.9.1 is incompatible with `protractor-hotkeys`. Is it necessary for yarn this version of node? Also I get another error https://travis-ci.org/inveniosoftware-contrib/ng2-json-editor/jobs/194531253#L154 . Do you know what to do for `yarn`?

jmartinm:
That is strange since we were before testing with Node 5 and the latest version of Node (7.x) and it was passing? Locally I also have 7.x installed. Now I changed to just test 6.9.1 and the latest version since 6.9.1 is a long time support version and is going to be around for the next years.

zzacharo:
For `protractor-hotkeys` it says that requires node <= 5.1. For yarn I pushed again to travis the linux environment to check if the problem  is specific to osx environment.

jmartinm:
yarn.lock is automatically updated, you need to do 'yarn install xxxx'  instead of 'npm install xxx' and the file should be updated automatically.

Or update package.json and then do the command 'yarn'  to install all dependencies and update yarn.lock

zzacharo:
On linux it fails again https://travis-ci.org/inveniosoftware-contrib/ng2-json-editor/jobs/194745778#L368

jmartinm:
Just checked and that package 'protractor-hotkeys'  seems quite dead so we should not rely on it (specially if it only works with old versions of Node). Can you check for alternatives?

zzacharo:
I will check it. For the `yarn` error do you have any idea?

jmartinm:
I don't see any other error besides the protractor-hotkeys version problem. That makes the yarn command fail. I am checking from the phone at the moment so maybe missed it

zzacharo:
@jmartinm Removed `protractor-hotkeys`. Done it with protractor pure methods like this:
```
elem.sendKeys(protractor.Key.chord(mod, protractor.Key.BACK_SPACE));
```

zzacharo:
@jmartinm @harunurhan I refactor the code and everything passed on travis. Also I created this issue #153 for investigating the `osx` travis configuration.

harunurhan:
@jmartinm I tested a bit and ran the `e2e` locally. Everything was OK. I didn't re-review the code but I think it's fine to merge.

PR #102, opened 2016-12-07
add title-dropdown (fixes #92)

### LINKED ISSUE #92: Improve complex field lists interface
The interface for complex list fields (such as references) is currently confusing due to lists nested inside of objects.

For example, in the image below there are two X buttons to delete an element from the array and the whole `misc` field.

image

In order to declutter the interface and as a bonus support https://github.com/inveniosoftware-contrib/ng2-json-editor/issues/7 I suggest to modify the complex list fields UI removing the add and delete buttons:

image

And instead making the label a [button-dropdown](http://getbootstrap.com/components/#btn-dropdowns):

editor

harunurhan:
* Adds title-dropdown that is used to hide some action buttons
such as add-new-element and delete.

* Removes extra space from button-holder column.

### DIFF
--- a/src/complex-list-field/complex-list-field.component.html
+++ b/src/complex-list-field/complex-list-field.component.html
@@ -31,8 +31,8 @@
         </label>
         <br>
         <pagination [totalItems]="values.size" [ngModel]="currentPage" [maxSize]="navigator.maxVisiblePageCount" [itemsPerPage]="navigator.itemsPerPage"
-          class="pagination-sm pagination-top" [boundaryLinks]="true" [rotate]="false" [firstText]="'❮❮'" [previousText]="'❮'" [nextText]="'❯'"
-          [lastText]="'❯❯'" (pageChanged)="onPageChange($event.page)"></pagination>
+          class="pagination-sm pagination-top" [boundaryLinks]="true" [rotate]="false" [firstText]="'❮❮'" [previousText]="'❮'"
+          [nextText]="'❯'" [lastText]="'❯❯'" (pageChanged)="onPageChange($event.page)"></pagination>
       </td>
     </tr>
   </table>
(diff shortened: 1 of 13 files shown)


src/complex-list-field/complex-list-field.component.html
```diff
@@ -42,8 +42,14 @@
         <tr *ngFor="let key of keys[i] | addAlwaysShowFields:schema.items | filterAndSortBySchema:schema.items; trackBy:trackByFunction">
           <td class="label-holder">
             <div>
-              <label>{{key | underscoreToSpace}}</label>
-              <add-new-element-button *ngIf="schema.items.properties[key].type === 'array'" [path]="getValuePath(pIndex, key)" [schema]="schema.items.properties[key]"></add-new-element-button>
+              <title-dropdown [title]="key | underscoreToSpace">
```
jmartinm: Can you make the style of this dropdown like the bootstrap one? (See that the whole row is clickable and has a background on hover)

dropdown

src/complex-list-field/complex-list-field.component.ts
```diff
@@ -100,10 +100,19 @@ export class ComplexListFieldComponent extends AbstractListFieldComponent implem
     }
   }
 
-  onFieldAdd(field: string, index: number) {
+  onFieldAdd(index: number, field: string) {
     this.keys[index].push(field);
   }
 
+  deleteField(index: number, field: string) {
+    // remove it the field from the element of list (record)
```
jmartinm: This comment is a bit difficult to understand as is, might be better without it.

src/complex-list-field/complex-list-field.component.ts
```diff
@@ -100,10 +100,19 @@ export class ComplexListFieldComponent extends AbstractListFieldComponent implem
...
+    this.values = this.values.removeIn([index, field]);
+    this.jsonStoreService.setIn(this.path, this.values);
+
+    // remove it from keys too, so that it will not be displayed as empty
```
jmartinm: This comment seems fine.

src/title-dropdown/title-dropdown.component.scss
```diff
@@ -0,0 +1,8 @@
+.title-dropdown-toggle {
```
jmartinm: See comment above about styling.

src/complex-list-field/complex-list-field.component.html
harunurhan: Actually this was my intention, and I had this on hover background but when I moved the html into `title-dropdown`, somehow it disappeared. However I'll look into it again.

harunurhan:
@jmartinm addressed all comments.

PR #106, opened 2016-12-08
add configurable link button to primitive-field

### LINKED ISSUE #80: ui: template next to field value
Some fields like DOI (look like  10.1088/1126-6708/2008/04/037) and arXiv ePrints (look like arXiv:1611.05765) should have a template near their value.

This will allow to have, for example, a link to the arXiv landing page or to the DOI landing page.

This should be generic since we might come up at any time with new content we want to add next to a field.

harunurhan:
* Adds x_editor_link_builder function that takes a primitive value
and returns a url string which will be pointed by a small anchor element
next to the input element of primitive field. (Fixes #80)

* Adds example config for x_editor_link_builder.

### DIFF
--- a/src/primitive-field/primitive-field.component.html
+++ b/src/primitive-field/primitive-field.component.html
@@ -1,30 +1,39 @@
 <div [ngSwitch]="valueType" [id]="path.join('.')">
-  <div class="editable-field-container" [ngClass]="errorNgClass" [tooltipHtml]="errors | errorsToMessagesHtml" tooltipTrigger="mouseenter"
-    [tooltipEnable]="isErrorTooltipEnabled">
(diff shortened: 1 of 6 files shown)


src/primitive-field/primitive-field.component.html
```diff
@@ -1,30 +1,39 @@
...
+            ## Not recognized type: {{valueType}}
+          </div>
+        </div>
+        <div>
+          <div *ngSwitchCase="'disabled'">
+            {{value}}
+          </div>
+        </div>
+      </td>
+      <td class="ink-button-holder">
```
jmartinm: `ink-button-holder` -> `link-button-holder`

harunurhan:
@jmartinm  fixed the typo

Issue #91, opened 2016-11-29
Tree menu overflowed


zzacharo:
![screenshot from 2016-11-29 16-23-56](https://cloud.githubusercontent.com/assets/22594783/20715775/87a11be0-b650-11e6-9be9-395fb5d2909b.png)

jmartinm:
@zzacharo can you add a description of what the problem is exactly?

zzacharo:
The tooltips and dropdown menu of the left items are cuurently be hidden from the tree-menu colum. Probably the css of the tree-menu column has a z-index higher than the middle coumn and thus the popup elements are hidden below.

jmartinm:
The same problem happened on the right side with the PDF panel. What I saw is that even if the `z-index` is correctly set the menus are still truncated.

The problem seems to be actually in the central div where the editor is. Since we have the scroll auto in the Y-axis this makes the content not show after the sides of the editor.

A workaround is to adjust the position of the messages and dropdowns on the left side so that they are positioned inside the editor.

zzacharo:
oh i see..

jmartinm:
The popover has been removed and the dropdowns are inside of the editor container, so this is not a problem any more.

PR #111, opened 2016-12-09
fix searchable dropdown alignment

### LINKED ISSUE #107: Enum dropdown not aligned with input
screen shot 2016-12-08 at 6 24 20 pm

It should take the same width as the textarea above and the text left aligned.

harunurhan:
* Fixes #107

screen shot 2016-12-09 at 17 00 54

### DIFF
--- a/src/searchable-dropdown/searchable-dropdown.component.scss
+++ b/src/searchable-dropdown/searchable-dropdown.component.scss
@@ -19,3 +19,7 @@ $placeholder-color: #000;
 div.btn-group {
     width: 100%;
 }
+
+.dropdown-menu {
+    left: 0px !important;
+}


harunurhan:
@jmartinm 

https://github.com/harunurhan/ng2-json-editor/blob/fix-107/src/json-editor.component.scss#L175

I have no idea what this does exactly and why it is necessary, but if you think it's OK to remove this class, we can remove it then add `right: 0px` instead of `left: 0px !important`. Or we could just remove `left: inherint` in `json-editor.component.scss` which will work exactly like this without touching `searchable-dropdown.component.scss`

jmartinm:
If I remember correctly, I added it to fix the issue similar to https://github.com/inveniosoftware-contrib/ng2-json-editor/issues/91 but on the right side. (When adding new colums to a table list the dropdown was getting truncated on the right side)

You can check if it still works without it.

harunurhan:
@jmartinm ok then this way is better, so it's OK to merge.

PR #114, opened 2016-12-13
history: Creates a json history functionality

zzacharo:


src/shared/services/json-store.service.spec.ts
```diff
@@ -86,4 +86,79 @@ describe('JsonStoreService', () => {
...
+            ]
+          }
+        ]
+      }
+    });
+    service.setJson(json);
+    let path = ['aMap', 'aList', 0, 'innerList', 0, 'innerProperty'];
+    for (let i = 1; i <= service.maxStateRestoreCounter; i++) {
+      service.addJsonToHistory();
+      service.setIn(path, `${i}`);
```
harunurhan: using `setJson` would be better not to depend on `setIn` in this test because it has its own tests.

src/shared/services/json-store.service.spec.ts
```diff
@@ -86,4 +86,79 @@ describe('JsonStoreService', () => {
     service.setIn(path, value);
   });
 
+  it('should shift the first element if the new pushed element exceeds the maximum history length', () => {
```
harunurhan: since the variables like `history` and `limit` are gonna be `private`, to test this:

* `setJson(initialJson)`
* `addHistory()`
* `setJson(anotherJson)`
* `addHistory()` x <limit> times
* `rollbackHistory()` x <limit> times (depending on the other comment you might need to change the json before each rollback)
*   `subscribe` and expect `changedJson` not equal to `initialJson`

zzacharo:
@harunurhan pushed changes.

src/add-new-element-button/add-new-element-button.component.ts
```diff
@@ -29,6 +29,7 @@ export class AddNewElementButtonComponent {
     let itemSchema = this.schema['items'];
     let emptyValue = this.emptyValueService.generateEmptyValue(itemSchema);
     let values: List<any> = this.jsonStoreService.getIn(this.path) || List();
+    this.jsonStoreService.addJsonToHistory();
```
harunurhan: I think this should be done after this PR and shortcut one.

src/abstract-list-field/abstract-list-field.component.ts
```diff
@@ -52,13 +52,15 @@ export abstract class AbstractListFieldComponent extends AbstractFieldComponent
...
       .set(newIndex, temp);
+    this.jsonStoreService.addJsonToHistory();
     this.jsonStoreService.setIn(this.path, this.values);
   }
 
   /**
    * @param {number} index - Index of the element to be deleted
    */
   deleteElement(index: number) {
+    this.jsonStoreService.addJsonToHistory();
```
harunurhan: I think this should be done after this PR and shortcut one.

src/shared/services/json-store.service.spec.ts
```diff
@@ -45,7 +46,7 @@ describe('JsonStoreService', () => {
         aList: [
           {
             innerList: [
-              { innerProperty: value }
+              {innerProperty: value}
```
harunurhan: unnecesary

src/shared/services/json-store.service.spec.ts
```diff
@@ -27,6 +27,7 @@ import { fromJS } from 'immutable';
 describe('JsonStoreService', () => {
 
   let service: JsonStoreService;
+  let historyLimit = 5;
```
harunurhan: in case we really need it, you can access it service['histroyLimit'];

zzacharo:
@harunurhan Made the changes. Check if there is something else.

src/shared/services/json-store.service.spec.ts
```diff
@@ -86,4 +86,73 @@ describe('JsonStoreService', () => {
...
+      anotherMap: {
+        anotherAnotherMap: {
+          anotherInnerMap: {
+            anotherInnerProperty: 'anotherValue'
+          }
+        }
+      }
+    });
+    service.setJson(anotherJson);
+
```
harunurhan: instead of hardcoding, `5`.
here 
`let historyLimit = service['historyLimit']`;

zzacharo:
@harunurhan changed `historyLimit`.

jmartinm:
Once the Travis error is fixed, should be good to go.

PR #117, opened 2016-12-14
add top-level add and delete field

### LINKED ISSUE #42: top level record should have add-field-dropdown
It should be on the left side, right above the `tree-menu`

### LINKED ISSUE #101: top level record should have delete button
Similar to #42

harunurhan:
* Fixes #101

* Fixes #42

* Refactors css regarding to margin and padding rules.

I did some extra styling, to make it more natural and understandable (at least to me 😄 ) but I am not sure if they are OK

@jmartinm

### DIFF
--- a/src/abstract-tracker/abstract-tracker.component.ts
+++ b/src/abstract-tracker/abstract-tracker.component.ts
@@ -22,13 +22,7 @@
 
 export abstract class AbstractTrackerComponent {
 
-  /**
-   * Used for track changes in model
-   * that is used to generate elements via *ngFor
-   * RELATED ISSUE: https://github.com/angular/angular/issues/4402
-   * TODO: Create a base component class that implements track by for all
-   */
-  trackByFunction(index: number, obj: any): any {
-    return index;
+  trackByFunction(index: number, element: any): any {
+    return element;
   }
 }
--- a/src/add-field-dropdown/add-field-dropdown.component.html
+++ b/src/add-field-dropdown/add-field-dropdown.component.html
@@ -1,12 +1,12 @@
 <div *ngIf="!disabled" class="btn-group" dropdown keyboardNav="true" (onToggle)="focusElementIfOpen($event, expressionInput)">
-  <button id="simple-btn-keyboard-nav" type="button" class="editor-btn-add-subfield" dropdownToggle>+ <span class="caret"></span></button>
+  <button id="simple-btn-keyboard-nav" type="button" class="btn-add-field-dropdown" dropdownToggle><ng-content></ng-content> <span class="caret"></span></button>
   <ul class="dropdown-menu" dropdownMenu aria-labelledby="simple-btn-keyboard-nav">
-    <li>
-      <input #expressionInput class="dropdown-filter" [(ngModel)]="expression" placeholder="filter">
+    <li class="dropdown-filter-container">
+      <input #expressionInput [(ngModel)]="expression" placeholder="filter">
     </li>
     <li class="divider dropdown-divider"></li>
     <li *ngFor="let key of schema | differentKeys:fields | filterByExpression:expression">
-      <a class="dropdown-item" href="javascript:void(0)" (click)="onFieldAdd.emit(key)">{{key}}</a>
+      <a class="dropdown-item" href="javascript:void(0)" (click)="onFieldSelect(key)">{{key}}</a>
     </li>
   </ul>
 </div>
--- a/src/complex-list-field/complex-list-field.component.html
+++ b/src/complex-list-field/complex-list-field.component.html
@@ -1,4 +1,4 @@
-<div [id]="path.join('.')">
+<div [id]="pathString">
   <!-- Navigator -->
   <table *ngIf="navigator" class="navigator-container">
     <tr>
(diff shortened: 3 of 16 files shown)


harunurhan:
@jmartinm 
Another question is: should we focus on new field when it is added?

src/abstract-tracker/abstract-tracker.component.ts
```diff
@@ -28,7 +28,7 @@ export abstract class AbstractTrackerComponent {
    * RELATED ISSUE: https://github.com/angular/angular/issues/4402
    * TODO: Create a base component class that implements track by for all
    */
-  trackByFunction(index: number, obj: any): any {
-    return index;
+  trackByFunction(index: number, element: any): any {
```
jmartinm: btw, do we still need the comment above? The issue https://github.com/angular/angular/issues/4402 is closed.

src/add-field-dropdown/add-field-dropdown.component.html
```diff
@@ -1,8 +1,8 @@
 <div *ngIf="!disabled" class="btn-group" dropdown keyboardNav="true" (onToggle)="focusElementIfOpen($event, expressionInput)">
-  <button id="simple-btn-keyboard-nav" type="button" class="editor-btn-add-subfield" dropdownToggle>+ <span class="caret"></span></button>
+  <button id="simple-btn-keyboard-nav" type="button" class="btn-add-field-dropdown" dropdownToggle><ng-content></ng-content> <span class="caret"></span></button>
```
jmartinm: Curiosity, What is the `<ng-content></ng-content>` for?

Not sure if it is due to these changes but the styling of the dropdown in table list has some small problems:

dropdown-filter

* The text inside the box should have the same dark color as the global `Add Field` you just added. Here is it very light.
* The button itself with the `+` to add a new element is too light and seems disabled. Should have the same color as that button is complex fields.

src/complex-list-field/complex-list-field.component.html
```diff
@@ -64,7 +64,7 @@
         <!-- ADD-FIELD-FROM-SCHEMA, UP/DOWN and DELETE buttons for each row group -->
         <tr *ngIf="values.size > 0">
           <td class="button-holder">
-            <add-field-dropdown [fields]="keys[i]" (onFieldAdd)="onFieldAdd(i, $event)" [schema]="schema.items.properties"></add-field-dropdown>
+            <add-field-dropdown [fields]="keys[i]" (onFieldAdd)="onFieldAdd(i, $event)" [schema]="schema.items.properties">+</add-field-dropdown>
```
jmartinm: I am not sure where the `+` is added in the UI. Cannot see the difference from before ;)

src/json-editor.component.ts
```diff
@@ -122,7 +122,19 @@ export class JsonEditorComponent extends AbstractTrackerComponent implements OnI
...
+    this.jsonStoreService.setJson(this._record);
+    // remove the key too, so that it will not be displayed as empty
+    this.keys.splice(this.keys.indexOf(name), 1);
+  }
+
   getPathForField(field: string): Array<any> {
     return [field];
   }
+
+  onFieldAdd(field: string) {
```
jmartinm: Above the parameter is called `name` and here `field`, are they the same thing? Can you a docstring on top explaining how the parameter looks like?

src/table-list-field/table-list-field.component.html
```diff
@@ -8,7 +8,7 @@
           </th>
 
           <th class="button-holder">
-            <add-field-dropdown *ngIf="values.size > 0" [fields]="keys" (onFieldAdd)="onFieldAdd($event)" [schema]="schema.items.properties"></add-field-dropdown>
+            <add-field-dropdown *ngIf="values.size > 0" [fields]="keys" (onFieldAdd)="onFieldAdd($event)" [schema]="schema.items.properties">+</add-field-dropdown>
```
jmartinm: Didn't we have the `+` sign already before?

jmartinm:
> Another question is: should we focus on new field when it is added?

Yes.

src/add-field-dropdown/add-field-dropdown.component.html
harunurhan: `<ng-content></ng-content>` is to make `add-field-dropdown` title dynamic, so that we can set it where we define the dropdown like below.

```
<add-field-dropdown...>Add a field</add-field-dropdown>
```

The same thing can be done via `@Input` but I felt like this is more natural.

src/complex-list-field/complex-list-field.component.html
harunurhan: the `+` that you see in the UI is actually this `+` since now we don't have it inside we pass it from here.

(see `<ng-content>` comment)

src/table-list-field/table-list-field.component.html
harunurhan: (see <ng-content> comment)

harunurhan:
@jmartinm  ready to check although it doesn't focus new added empty `object-field` and `table-list` but after #100 is merged it will focus the `add-field-dropdown` if the focused field doesn't have any field.

PR #118, opened 2016-12-14
Collectionsjs

jmartinm:


jmartinm:
This makes `SortedMap` from collections.js work.

The only 'small' thing is that typescript compiler gives the warning:

```javascript
ERROR in [default] /Users/jmartinm/.virtualenvs/invenio3/src/editor/ng2-json-editor/vendor.ts:1:27
Cannot find module 'collections/sorted-map'
```

And to make it happy we need to add to `example/typings.d.ts`:

```javascript
declare module 'collections/sorted-map';
```

Adding this line to the `src/typings.d.ts` does not work. Not sure if there is some way without modifying `example/` folder.

jmartinm:
Converted into an issue https://github.com/inveniosoftware-contrib/ng2-json-editor/issues/131

PR #121, opened 2016-12-19
Karma chrome fix

jmartinm:


karma.conf.js
```diff
@@ -34,6 +34,9 @@ module.exports = function (config) {
     logLevel: config.LOG_INFO,
     autoWatch: true,
     browsers: process.env.TRAVIS ? ['PhantomJS'] : ['Chrome'],
-    singleRun: false
+    singleRun: false,
+    mime: {
+      'text/x-typescript': ['ts', 'tsx']
```
harunurhan: not important but, could be only `ts` since `tsx` is for `react` projects.

PR #122, opened 2016-12-19
move test browser to PhantomJS

harunurhan:
* Changes test browser for local runs to PhantomJS since local chrome
update sometimes breaks local test environment.


harunurhan:
Not needed anymore

PR #125, opened 2016-12-19
WIP decorators: Addition of decorator hook

zzacharo:


src/shared/services/app-globals.service.ts
```diff
@@ -35,4 +36,9 @@ export class AppGlobalsService {
     return this._globalErrorsSubject;
   }
 
+  @registerOnCommitHook()
```
jmartinm: So by adding the code above https://github.com/inveniosoftware-contrib/ng2-json-editor/pull/125/files#diff-14144842719ab4361ae218ed87c6613eR25

does it check the types for `checkSeparator(value: string, path: string)` ? to make sure `value` and `path` have the correct types or that did not work?

jmartinm:
Once we have a way that works for the hooks, separate into 2 commits. One introducing the hook service and another one using it for the input splitting.

harunurhan:
We will have look at this later.

PR #127, opened 2016-12-19
fix coverage percentage calculation

harunurhan:


harunurhan:
I have tried different combinations with this config but doesn't work @jmartinm.

PR #134, opened 2017-01-10
implement find-replace to resolve #103

### LINKED ISSUE #103: field scope find and replace support

harunurhan:
* Adds find-replace-all service

* Adds find replace for top level fields

* Adds modal-view component and service

### DIFF
--- a/src/shared/services/find-replace-all.service.ts
+++ b/src/shared/services/find-replace-all.service.ts
@@ -1,0 +1,59 @@
+import { Injectable } from '@angular/core';
+import { List, Map } from 'immutable';
+
+@Injectable()
+export class FindReplaceAllService {
+
+  // TODO: fix compiler errors when type of immutable `List<any> | Map<string, any>`
+  /**
+   * Does deep replace when it finds the occurance in an immutable List and Map
+   * uses schema to skip disabled properties
+   * 
+   * @param {List<any> | Map<string, any} immutable
+   * @param {Object} schema
+   * @param {string} find
+   * @param {replace} replace
+   * @param {boolean} matchWhole - looks for whole match to a string property of immutable
+   * @param {Array<any> | Object} diffHtml - TODO: describe
+   */
+  findReplaceInImmutable(immutable: any, schema: Object, find: string, replace: string, matchWhole = false, diffHtml?: Array<any> | Object):
+    { replaced: any, diffHtml: Array<any> | Object } {
+    let immutableAsMutable = immutable.asMutable();
+    let isList = List.isList(immutable);
+    // create empty array or object for the immutable to store diff
+    diffHtml = isList ? [] : {};
+    immutableAsMutable.forEach((value, key) => {
+      let innerSchema = isList ? schema['items'] : schema['properties'][key];
+      // ignore disabled and ref fields
+      if (innerSchema['x_editor_disabled'] || innerSchema['x_editor_hidden'] || key === '$ref') {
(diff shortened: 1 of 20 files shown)


src/typings.d.ts
```diff
@@ -80,6 +80,13 @@ interface Preview {
   getUrl?: GetUrlFunction;
 }
 
+interface ModalOptions {
+  title: string;
+  bodyHtml: any; // TODO: change to SafeHtml
```
harunurhan: importing `SafeHtml` breaks the code.

jmartinm:
On the UI side, the dropdown titles and elements are not aligned. They should look like in the example http://getbootstrap.com/components/#dropdowns-headers

Then the `wholeMatch` checkbox is not obvious. I would prefer to see a toggle icon (with a tooltip on hover). Something like:

whole-toggle

src/shared/services/find-replace-all.service.ts
```diff
@@ -0,0 +1,42 @@
...
+    { replaced: any, diffHtml: Array<any> | Object } {
+    let immutableAsMutable = immutable.asMutable();
+    // create empty array or object for the immutable to store store diff
+    diffHtml = List.isList(immutable) ? [] : {};
+    immutableAsMutable.forEach((value, key) => {
+      if (typeof value === 'string') {
+        // assign value to diff as initial, if nothing has changed it will remain same
+        let diff = value;
+        // create html diff for each possible change
+        let singleDiffHtml = `<strong style='color: green;'>${replace}</strong><del><em style='color: red;'>${find}</em></del>`;
```
harunurhan: I was thinking the same but then I had to add this class to somewhere global, so that's why I wasn't sure to add it or hardcode it in string.

jmartinm:
There seems to be a bug with the dropdown not closing in this case:

1) Open the editor
2) Go to authors -> affiliations and click `Add New`
3) Use the dropdown to add a new field to the author (emails and so on)

This dropdown stays open while clicking on items. (it only seems to happen when doing things in the order above)

jmartinm:
A similar thing happens in any field (in top level title dropdown)

1) Open dropdown and click `Add New`
2) Open dropdown again and now try to click outside of the dropdown (it does not close)

jmartinm:
Looking a the dropdown, I think at the moment the titles are a bit distracting and do not bring much information, so I would remove them and leave it plain like this:

screen shot 2017-01-19 at 4 22 09 pm

(for the divider I have used http://getbootstrap.com/components/#dropdowns-divider)

jmartinm:
Match whole in the dropdown is not working. When selecting it, no replacement is done.

harunurhan:
@jmartinm comments addressed.

jmartinm:
There is one corner case not covered. Find/Replace will replace content even if a field is set as non_editable so this might have some unwanted consequence.

harunurhan:
@jmartinm it skips disabled objects, arrays, string fields now, all cases are covered by the test as well.

jmartinm:
Can you double check? For me it does not seem to work properly:


find-replace

harunurhan:
@jmartinm I only ignored `x_editor_disabled: true` fields. Totally forgot about `$ref` fields. I'll exclude `$ref` as well but do we have something else?

jmartinm:
Are hidden fields excluded?

harunurhan:
@jmartinm now hidden ones are excluded too, also rebased!

PR #140, opened 2017-01-12
purify pipes and use immutables in pipes

### LINKED ISSUE #89: Purify pipes
We should change `pure` state of our pipes from `false` to default which is `true`, unless a pipe is using other arguments than `@Input()` properties.

Might be helpful: https://angular.io/docs/ts/latest/guide/pipes.html

harunurhan:
* Fixes #89

* Purifies pipes that take primitive arguments.

* Changes pipe arguments from Array to Immutable.Set
and purify those pipes.

* Changes `keys` type in components from Array to Immutable.Set
so that it can be passed to pipes.

### DIFF
--- a/src/add-field-dropdown/add-field-dropdown.component.spec.ts
+++ b/src/add-field-dropdown/add-field-dropdown.component.spec.ts
@@ -22,6 +22,8 @@
 
 import { async, ComponentFixture, TestBed } from '@angular/core/testing';
 
+import { Set } from 'immutable';
+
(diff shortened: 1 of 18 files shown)


jmartinm:
Can you rebase with latest master?

zzacharo:
@harunurhan Can you push again to test it locally? For the warnings you fixed?

harunurhan:
@jmartinm rebased
@zzacharo pushed the new one that doesn't have any compile warnings

PR #142, opened 2017-01-13
fix-issue-115: Disable buttons and dropdowns along with fields

### LINKED ISSUE #115: Disable field when x_editor_disabled: true
When a field is marked with `x_editor_disabled: true` the field should have all the buttons disabled (the top level Dropdown to Add/Delete) and the internal move/delete buttons.

zzacharo:
(Closes #115)

### DIFF
--- a/example/app/app.config.ts
+++ b/example/app/app.config.ts
@@ -19,6 +19,9 @@ export const EXAMPLE_CONFIG: AppConfig = {
         maxVisiblePageCount: 5
       }
     },
+    'license': {
+      x_editor_disabled: true
+    },
     'arxiv_eprints.items.properties.value': {
       x_editor_link_builder: (value: any) => {
         return `http://arxiv.org/abs/${value}`;
--- a/src/any-type-field/any-type-field.component.html
+++ b/src/any-type-field/any-type-field.component.html
@@ -1,23 +1,25 @@
-<div [ngSwitch]="componentType">
-  <div *ngSwitchCase="'table-list'">
-    <table-list-field [values]="value" [schema]="schema" [path]="path"></table-list-field>
-  </div>
-  <div *ngSwitchCase="'complex-list'">
-    <complex-list-field [values]="value" [schema]="schema" [path]="path"></complex-list-field>
-  </div>
-  <div *ngSwitchCase="'primitive-list'">
-    <primitive-list-field [values]="value" [schema]="schema" [path]="path"></primitive-list-field>
-  </div>
-  <div *ngSwitchCase="'object'">
-    <object-field [value]="value" [schema]="schema" [path]="path"></object-field>
-  </div>
-  <div *ngSwitchCase="'ref'">
-    <ref-field [value]="value" [schema]="schema" [path]="path"></ref-field>
-  </div>
-  <div *ngSwitchCase="'value-change-watcher'">
-    <value-change-watcher [value]="value" [schema]="schema" [path]="path"></value-change-watcher>
-  </div>
-  <div *ngSwitchDefault>
-    <primitive-field [value]="value" [schema]="schema" [path]="path"></primitive-field>
+<div [ngClass]="{disableCursorClass: schema && schema.x_editor_disabled}">
+  <div [ngSwitch]="componentType" [ngClass]="{disableEventsClass: schema && schema.x_editor_disabled}">
+    <div *ngSwitchCase="'table-list'">
+      <table-list-field [values]="value" [schema]="schema" [path]="path"></table-list-field>
+    </div>
+    <div *ngSwitchCase="'complex-list'">
+      <complex-list-field [values]="value" [schema]="schema" [path]="path"></complex-list-field>
+    </div>
+    <div *ngSwitchCase="'primitive-list'">
+      <primitive-list-field [values]="value" [schema]="schema" [path]="path"></primitive-list-field>
+    </div>
+    <div *ngSwitchCase="'object'">
+      <object-field [value]="value" [schema]="schema" [path]="path"></object-field>
+    </div>
+    <div *ngSwitchCase="'ref'">
+      <ref-field [value]="value" [schema]="schema" [path]="path"></ref-field>
+    </div>
+    <div *ngSwitchCase="'value-change-watcher'">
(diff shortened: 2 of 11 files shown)


example/app/app.config.ts
```diff
@@ -19,6 +19,9 @@ export const EXAMPLE_CONFIG: AppConfig = {
         maxVisiblePageCount: 5
       }
     },
+    'accelerator_experiments': {
```
jmartinm: Better to disable a 'less' interesting field, such as license.

jmartinm: Better to disable a 'less' interesting field, such as license.

src/any-type-field/any-type-field.component.html
```diff
@@ -1,23 +1,23 @@
-<div [ngSwitch]="componentType">
-  <div *ngSwitchCase="'table-list'">
+<div [ngSwitch]="componentType" [ngClass]="cursorDisabled">
```
jmartinm: [ngClass]="{'disableCursor': schema['x_editor_disabled']}"

src/json-editor.component.scss
```diff
@@ -215,4 +215,18 @@ li.title-dropdown-item > a {
       color: black;
     }
   }
+}
+
+.disableCursor {
+  cursor: not-allowed;
+}
+
+.disableEventsClass {
```
jmartinm: duplicate

src/any-type-field/any-type-field.component.html
```diff
@@ -1,23 +1,23 @@
-<div [ngSwitch]="componentType">
-  <div *ngSwitchCase="'table-list'">
+<div [ngSwitch]="componentType" [ngClass]="{disableCursorClass: schema && schema['x_editor_disabled'] ? true : false}">
```
harunurhan: Isn't this enough since this is the parent div for the field which is disabled.

zzacharo: @harunurhan What do you mean?

zzacharo: @harunurhan The `disableCursorClass` set cursor to not-allowed. The `disableEventsClass` is actually disabling the pointer-events. I have it both because if apply the both classes to the same div the cursor is the default one because there is a known conflict when applying pointer-events:none and cursor:not-allowed. The solution was to apply it to nested divs. If we don't want the cursor thing I can apply only the pointer-events:none css to the div you mentioned.

src/any-type-field/any-type-field.component.html
```diff
@@ -1,23 +1,25 @@
...
-  </div>
-  <div *ngSwitchCase="'ref'">
-    <ref-field [value]="value" [schema]="schema" [path]="path"></ref-field>
-  </div>
-  <div *ngSwitchCase="'value-change-watcher'">
-    <value-change-watcher [value]="value" [schema]="schema" [path]="path"></value-change-watcher>
-  </div>
-  <div *ngSwitchDefault>
-    <primitive-field [value]="value" [schema]="schema" [path]="path"></primitive-field>
+<div [ngClass]="{disableCursorClass: schema && schema['x_editor_disabled'] ? true : false}">
```
jmartinm: Not enough with 

```
schema && schema['x_editor_disabled']
```
?

zzacharo:
@jmartinm yes was a leftover. I pushed again the code.

src/any-type-field/any-type-field.component.html
```diff
@@ -1,23 +1,25 @@
...
-  </div>
-  <div *ngSwitchCase="'ref'">
-    <ref-field [value]="value" [schema]="schema" [path]="path"></ref-field>
-  </div>
-  <div *ngSwitchCase="'value-change-watcher'">
-    <value-change-watcher [value]="value" [schema]="schema" [path]="path"></value-change-watcher>
-  </div>
-  <div *ngSwitchDefault>
-    <primitive-field [value]="value" [schema]="schema" [path]="path"></primitive-field>
+<div [ngClass]="{disableCursorClass: schema && schema['x_editor_disabled']}">
```
harunurhan: `schema.x_editor_disabled` is fine in the template, you don't have to put brackets.

src/any-type-field/any-type-field.component.html
```diff
@@ -1,23 +1,25 @@
...
+  <div [ngSwitch]="componentType" [ngClass]="{disableEventsClass: schema && schema['x_editor_disabled']}">
```
harunurhan: ditto

src/title-dropdown/title-dropdown.component.ts
```diff
@@ -32,4 +32,5 @@ import { Component, Input, ChangeDetectionStrategy } from '@angular/core';
 })
 export class TitleDropdownComponent {
   @Input() title: string;
+  @Input() schema: Object;
```
harunurhan: Instead of this we can have an input `isDisabled: boolean`.

src/title-dropdown/title-dropdown.component.html
```diff
@@ -1,7 +1,9 @@
-<div dropdown>
-  <button class="btn title-dropdown-toggle" dropdownToggle>
+<div dropdown [ngClass]="{disableCursorClass: schema && schema['x_editor_disabled']}">
+  <div [ngClass]="{disableEventsClass: schema && schema['x_editor_disabled']}">
+    <button class="btn title-dropdown-toggle" dropdownToggle>
```
harunurhan: instead of `disableEventsClass` you can use `[isDisabled]` of `dropdownToggle` which will do the necessary thing, then you can remove the extra div with `[ngClass]`. Also I think you can even remove the `disableCursorClass`, but I am not 100% sure that `[isDisabled]` will handle this.

zzacharo: but how i will know if it is disabled? I mean i ll pass the schema and apply the dropdown isDisabled instead of the class you mean?

zzacharo: @harunurhan It worked but for the title-dropdown when you on hover it changes the color but the cursor is not-allowed. Check it again from the latest commit.

PR #143, opened 2017-01-18
fix-issue-137: Fix issue with empty primitive arrays

### LINKED ISSUE #137: Add empty placeholder in Array for table list
When opening an author record in the editor, the `positions` field looks like:

screen shot 2017-01-12 at 11 12 00 am

When a row does not have emails it looks confusing with a double `Add New`. That can be improved if we add a placeholder empty cell when there is no email.

zzacharo:
* Generates an empty primitive value if array is empty.
(Closes #137)

### DIFF
--- a/src/shared/pipes/self-or-empty.pipe.spec.ts
+++ b/src/shared/pipes/self-or-empty.pipe.spec.ts
@@ -1,0 +1,99 @@
+/*
+ * This file is part of ng2-json-editor.
+ * Copyright (C) 2017 CERN.
+ *
+ * ng2-json-editor is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License as
+ * published by the Free Software Foundation; either version 2 of the
+ * License, or (at your option) any later version.
+ *
+ * ng2-json-editor is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with ng2-json-editor; if not, write to the Free Software Foundation, Inc.,
+ * 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
+ * In applying this license, CERN does not
+ * waive the privileges and immunities granted to it by virtue of its status
+ * as an Intergovernmental Organization or submit itself to any jurisdiction.
+ */
+
+import { EmptyValueService } from '../services/empty-value.service';
+import { SelfOrEmptyPipe } from './self-or-empty.pipe';
+import { Map, List } from 'immutable';
+
+describe('SelfOrEmptyPipe', () => {
+  let pipe: SelfOrEmptyPipe;
+
+  beforeEach(() => {
+      pipe = new SelfOrEmptyPipe(new EmptyValueService());
+  });
+
+  it('should return empty value if given value is undefined', () => {
+    let schema = {
+      type: 'object',
+        properties: {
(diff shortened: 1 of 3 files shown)


src/shared/pipes/self-or-empty.pipe.spec.ts
```diff
@@ -0,0 +1,79 @@
...
+ * In applying this license, CERN does not
+ * waive the privileges and immunities granted to it by virtue of its status
+ * as an Intergovernmental Organization or submit itself to any jurisdiction.
+ */
+
+import { EmptyValueService } from '../services/empty-value.service';
+import { SelfOrEmptyPipe } from './self-or-empty.pipe';
+import {fromJS, List} from 'immutable';
+
+describe('SelfOrEmptyPipe', () => {
```
harunurhan: tab size must be 2 spaces, unfortunately there is no `tslint` config for the tabsize yet that's why we can't enforce it.

src/shared/pipes/self-or-empty.pipe.spec.ts
```diff
@@ -0,0 +1,79 @@
...
+
+    it('should return empty value if given value is immutable List and is empty', () => {
+        let schema = {
+            type: 'array',
+            items: {
+                type: 'string'
+            }
+        };
+
+        let value: List = fromJS([]);
```
harunurhan: for creating immutables with simple javascript object please use constructors instead of `fromJS` 
Example: `let value = List<WhateverType>();`

After this update it will be fixed but this has actually syntax error, you can use `List` without giving element type.

src/shared/pipes/self-or-empty.pipe.spec.ts
```diff
@@ -0,0 +1,79 @@
...
+                }
+            }
+        };
+
+        let value = undefined;
+        let pipeResult = pipe.transform(value, schema);
+        expect(pipeResult).toBeDefined();
+    });
+
+    it('should return empty value if given value is immutable List and is empty', () => {
```
harunurhan: can you add something like this, but when the array type is `object` with a property.

src/shared/pipes/self-or-empty.pipe.spec.ts
```diff
@@ -0,0 +1,79 @@
+/*
+ * This file is part of ng2-json-editor.
+ * Copyright (C) 2016 CERN.
```
jmartinm: 2017 ;)

src/shared/pipes/self-or-empty.pipe.spec.ts
```diff
@@ -0,0 +1,79 @@
...
+            properties: {
+                prop1: {
+                    type: 'string'
+                }
+            }
+        };
+
+        let value = undefined;
+        let pipeResult = pipe.transform(value, schema);
+        expect(pipeResult).toBeDefined();
```
jmartinm: Can you add explicitly what there should be in `pipeResult` ? Just by reading the test I can see it should be defined but not sure with which content.

jmartinm:
> Addition to my comments, I am not sure if we really need to author schema and record to the repo?

Let's remove it from here and we can create protractor tests in http://github.com/inspirehep/record-editor that will open different sample records and make sure they display as we want.

src/shared/pipes/self-or-empty.pipe.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+  it('should return empty value if given value is immutable List and is empty', () => {
+    let schema = {
+      type: 'array',
+      items: {
+        type: 'string'
+      }
+    };
+
+    let value = List();
+    let expectedResult = List<any>(['']);
```
harunurhan: `List([''])` or ` List<string>([''])` which is better.

PR #151, opened 2017-01-23
add admin mode

### LINKED ISSUE #40: Support enable & disable advance/admin mode
There should be a check button to enable/disable advance/admin mode in which it is possible to see `x_editor_hidden` and edit `x_editor_disabled` fields.

harunurhan:
* Fixes #40 

UI to be discussed.

admin-mode

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,7 @@ Please have a look at [example app](./example/app)
 | *schemaOptions*      | Object | Enriches JsonSchema with custom properties that `ng2-json-editor` understands. See [possible values](#schemaOptions).|
 | *previews*      | Array | Configures previews for the document. See [configuration syntax](#previews).|
 | *shortcuts*     | Object | Configures shortcuts for the document. See [shortcuts syntax](#shortcuts).|
+| *enableAdminModeSwitch*      | boolean=false | If set, displays the switch that can enable admin mode where any field can be edited.|
 
 ### <a name="schemaOptions"></a>Extended Schema (schemaOptions)
 
--- a/src/json-editor.component.html
+++ b/src/json-editor.component.html
@@ -2,6 +2,10 @@
   <div class="col-md-2 left">
     <tree-menu [record]="_record" [schema]="schema"></tree-menu>
     <add-field-dropdown [fields]="keys" [pathString]="''" (onFieldAdd)="onFieldAdd($event)" [schema]="schema.properties">Add field</add-field-dropdown>
+    <div *ngIf="config.enableAdminModeSwitch" class="admin-mode" tooltip="Allows editing all fields (use with care)">
+      <input id="admin-mode-checkbox" type="checkbox" [(ngModel)]="appGlobalsService.adminMode" />
+      <label class="admin-mode" for="admin-mode-checkbox">Enable Admin Mode</label>
+    </div>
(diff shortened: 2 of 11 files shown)


jmartinm:
LGTM.

For the UI, I would for the moment leave it as you have it with a label next to the checkbox. `Unlock protected fields`

harunurhan:
Updated:

screen shot 2017-01-23 at 17 41 22

jmartinm:
The issue I have with `Enable Admin Mode` is that it is not self explanatory, you will have to click to see what happens.

We can leave it like that but add a tooltip on hover that says: "Allows editing all fields (use with care)".

zzacharo:
This LGTM too. Also maybe the ui of this admin could be part of the record-editor on the toolbar. So, we could have a global exported admin input on our json-editor app.

jmartinm:
> This LGTM too. Also maybe the ui of this admin could be part of the record-editor on the toolbar. So, we could have a global exported admin input on our json-editor app. And have some flag like cancel_hidden_disable_filters here instead of adminMode.

Yes, that's a good idea since this is something quite INSPIRE-specific and probably another installation using the editor would not want to allow people to edit everything.

Can we have a config variable that enables/disables this feature and we set it to `true` on the record-editor?

jmartinm:
Also for it to be ready note [the change](https://github.com/inveniosoftware-contrib/ng2-json-editor/pull/151#issuecomment-274545785) suggested above.

harunurhan:
@zzacharo I think it should be here at least for now, but we can have a config to disable admin mode which will never display the checkbox.

@jmartinm addressed your change

harunurhan:
Now your comment, so will add the config.

harunurhan:
@jmartinm added the config, and rebased it.

Also realized that we should really stop using typings because for config type, it needs be repeated for example.

README.md
```diff
@@ -43,6 +43,7 @@ Please have a look at [example app](./example/app)
 | *schemaOptions*      | Object | Enriches JsonSchema with custom properties that `ng2-json-editor` understands. See [possible values](#schemaOptions).|
 | *previews*      | Array | Configures previews for the document. See [configuration syntax](#previews).|
 | *shortcuts*     | Object | Configures shortcuts for the document. See [shortcuts syntax](#shortcuts).|
+| *enableAdminModeSwitch*      | boolean=false | If set, displays the switch that can enable admin mode where anything field can be edited.|
```
jmartinm: `anything field` -> `any field`

src/json-editor.component.html
```diff
@@ -2,6 +2,10 @@
   <div class="col-md-2 left">
     <tree-menu [record]="_record" [schema]="schema"></tree-menu>
     <add-field-dropdown [fields]="keys" [pathString]="''" (onFieldAdd)="onFieldAdd($event)" [schema]="schema.properties">Add field</add-field-dropdown>
+    <div *ngIf="config.enableAdminModeSwitch"class="admin-mode" tooltip="Allows editing all fields (use with care)">
```
jmartinm: space missing after `*ngIf="config.enableAdminModeSwitch"`

harunurhan:
@jmartinm addressed the comments

PR #157, opened 2017-01-25
issue-152: Add link button to editable links

### LINKED ISSUE #152: Clickable URLs
When a JSON document has URLs in it, they should be clickable.

screen shot 2017-01-24 at 10 01 50 am

The ones in $ref are, but not the other ones.

* An important thing is that for editable fields with a URL inside it needs to be possible to click on them to edit, so it can be annoying if the whole text is a link. So the best solution can be to be consistent with the current way we display links in identifiers

screen shot 2017-01-24 at 10 04 25 am

And also show such a link icon close to the fields with a URL on them.

zzacharo:
* Adds the link button to editable fields with format:'url' in their schema.
(Closes #152)

### DIFF
--- a/src/primitive-field/primitive-field.component.html
+++ b/src/primitive-field/primitive-field.component.html
@@ -28,6 +28,7 @@
       </td>
       <td class="link-button-container">
         <a *ngIf="schema.linkBuilder" class="no-decoration" target="_blank" [href]="schema.linkBuilder(value)">🔗</a>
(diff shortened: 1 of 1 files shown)


harunurhan:
I changed this part quite in the #159, we can merge this but I will have to kind of rewrite it anyway.

zzacharo:
@harunurhan check this also!

PR #159, opened 2017-01-26
add type support for config and refactoring

harunurhan:
* Adds type support for configuration.

* Removes typings.d.ts.

* Refactor names of some schema options.

* Removes `x_editor_` prefix from schemaOptions.


src/shared/interfaces/nested-store.ts
```diff
@@ -0,0 +1,4 @@
+export interface NestedStore {
+  getIn(path: Array<any>): any;
+  getIn(path: Array<any>): any;
```
jmartinm: `setIn` ?

jmartinm:
Works fine. The only thing is that on `npm start` there are a few warnings:

```
src/autocomplete-input/autocomplete-input.component.ts(89,61): error TS2345: Argument of type 'JsonStoreService' is not assignable to parameter of type 'NestedStore'.
  Types of property 'setIn' are incompatible.
    Type '(path: any[], value: any) => void' is not assignable to type '(path: any[]) => any'.
src/shared/services/json-store.service.ts(8,14): error TS2420: Class 'JsonStoreService' incorrectly implements interface 'NestedStore'.
  Types of property 'setIn' are incompatible.
    Type '(path: any[], value: any) => void' is not assignable to type '(path: any[]) => any'.
src/value-change-watcher/value-change-watcher.component.ts(74,71): error TS2345: Argument of type 'JsonStoreService' is not assignable to parameter of type 'NestedStore'.
```

jmartinm:
Superseded by https://github.com/inveniosoftware-contrib/ng2-json-editor/pull/167

PR #160, opened 2017-01-26
fix #154

### LINKED ISSUE #154: Clean primitive-field template
After the merge of #142 we can clean primitive-field.component.html and remove the default case of ngSwitch. After clearing we must refactor the `dom-util.service.ts` `focusAndSelectFirstInputChildById` method to exclude `input[type='boolean']` in order not to focus on checkboxes. This will be consistent with #100 misbehaviour, when you click on a checkbox and then tab the tabindexes order is not followed. 

The issue is related to chrome as shown here:
http://stackoverflow.com/questions/6761169
http://stackoverflow.com/questions/18400190

Maybe we can investigate how to fix this because I tried to explicitly set focus on checkbox on click but the click event was triggering after setting focus on it and the focus was cleared away.

harunurhan:
Just cleaned the template and some scss, check in case but shouldn't break anything

### DIFF
--- a/src/primitive-field/primitive-field.component.html
+++ b/src/primitive-field/primitive-field.component.html
@@ -1,37 +1,33 @@
 <div [ngSwitch]="valueType" [id]="pathString">
-  <table class="input-container">
+  <table class="primitive-field-container">
     <tr>
-      <td class="value-holder">
-        <div class="editable-field-container" [ngClass]="errorNgClass" [tooltipHtml]="errors | errorsToMessagesHtml" tooltipTrigger="mouseenter"
-             [tooltipEnable]="isErrorTooltipEnabled">
-          <div *ngSwitchCase="'string'">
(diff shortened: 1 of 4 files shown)


src/primitive-field/primitive-field.component.html
```diff
@@ -1,37 +1,34 @@
...
+        <div *ngSwitchCase="'string'">
+          <textarea rows="1" textareaAutofit [tabindex]="tabIndex" [(ngModel)]="value" (blur)="commitValueChange()" (keypress)="onKeypress($event)"
+            placeholder="{{schema.title}}"></textarea>
         </div>
-        <div>
-          <div *ngSwitchCase="'disabled'">
-            {{value}}
-          </div>
+        <div *ngSwitchCase="'enum'">
+          <!-- TODO: set placeholder -->
```
jmartinm: What was this comment about?

jmartinm: I can see its about `[placeholder]` is there an issue for this instead of having the comment here?

harunurhan: This was long time ago but still valid because we still have some hacky things in `searchable-dropdown` that makes use of `placeholder`. So we can't just set it.

However I agree I can create an issue, and remove this comment well.

PR #166, opened 2017-01-30
WIP different-records: E2e tests for different record types

zzacharo:


zzacharo:
@jmartinm We can check together the first tests for every document. RIght now I check for every document 3 tests each using some shortcuts on different fields. We can discuss about what else we need for now to test.

zzacharo:
@jmartinm take a look to this one. Is this what we want?

example/app/app-editor.component.ts
```diff
@@ -0,0 +1,78 @@
...
+        this.http.get(this.mockDataFactoryService.getRecordPathForRecordType(recordType)),
+        this.http.get(this.mockDataFactoryService.getSchemaPathForRecordType(recordType)),
+        (recordRes, schemaRes) => {
+          return {
+            record: recordRes.json(),
+            schema: schemaRes.json()
+          };
+        }
+      ).subscribe((data) => {
+        this.record = data.record; // set ./assets/mock-data/record.json
```
jmartinm: The comment does not seem helpful since you have different mock records now.

example/app/app.routing.ts
```diff
@@ -0,0 +1,14 @@
...
+import { Routes, RouterModule } from '@angular/router';
+import { AppEditorComponent } from './app-editor.component';
+
+
+const appRoutes: Routes = [
+  { path: '', component: AppEditorComponent },
+  { path: ':recordType', component: AppEditorComponent }
+];
+
+export const appRoutingProviders: Array<any> = [];
```
jmartinm: Do you know what is this used for?

jmartinm:
LGTM, @harunurhan can you have a look as well?

PS. It also seems like Travis did not run on latest version.

zzacharo:
@harunurhan @jmartinm You can review this again.

e2e/record-types/record-types.e2e-spec.ts
```diff
@@ -0,0 +1,57 @@
...
+ * In applying this license, CERN does not
+ * waive the privileges and immunities granted to it by virtue of its status
+ * as an Intergovernmental Organization or submit itself to any jurisdiction.
+ */
+
+import { RecordTypesPage } from './record-types.po';
+
+describe('Record-Types App', function() {
+  let page: RecordTypesPage;
+  let recordFile: Object;
```
jmartinm: This does not seem to be a File so name is a bit confusing, isn't just the record?

e2e/record-types/record-types.e2e-spec.ts
```diff
@@ -0,0 +1,57 @@
...
+  beforeEach(() => {
+    page = new RecordTypesPage();
+  });
+
+  it('should match the value of a random path in each record with the value of the same path in UI.', () => {
+    for (let index = 0; index < records.length; index++) {
+      let recordType = records[index];
+      recordFile = require(page.getPathForRecordType(recordType));
+      page.navigateTo(`/${recordType}`);
+      let idPath = page.getRandomIdPathFromObject(recordFile);
```
jmartinm: Since this function returns the id and the path (and not only the id) the name could be better. You can check http://www.2ality.com/2014/06/es6-multiple-return-values.html and this should work

```javascript
let {path, value} = page.getRandomIdPathAndValueFromObject(record);
```

e2e/record-types/record-types.e2e-spec.ts
```diff
@@ -0,0 +1,57 @@
...
+      let element = page.getChildOfElementByCss(page.getElementById(idPath.path), `textarea, input, a`);
+
+
+      expect(idPath.path).not.toEqual('');
```
jmartinm: This seems unrelated to this test? It looks like you are testing here your function `getRandomIdPathFromObject `

e2e/record-types/record-types.e2e-spec.ts
```diff
@@ -0,0 +1,57 @@
...
+      // convert false value to null to be compatible with protractor.getAttibute('checked') method
+      // as it returns null if checkbox is not checked.
+      expect(page.getElementValueByElement(element)).toEqual(page.convertIfFalseValueToNull(idPath.value));
```
jmartinm: `castFalseToNull` instead of `convertIfFalseValueToNull`

e2e/record-types/record-types.po.ts
```diff
@@ -0,0 +1,100 @@
...
+      return `${RecordTypesPage.VALUE_SEPARATOR}${item}`;
+    }
+  }
+
+  private pickRandomIndex(list: Array<any>) {
+    let randomIndex = Math.floor(Math.random() * list.length);
+    return randomIndex;
+  }
+
+  convertIfFalseValueToNull(value: string) {
```
jmartinm: should this be `private` like the one above?

zzacharo: No because I access it from the spec file it will not be visible if I make it private.

e2e/record-types/record-types.e2e-spec.ts
zzacharo: I deleted yes!

zzacharo:
@jmartinm made the changes!

gulpfile.js
```diff
@@ -113,6 +114,7 @@ gulp.task('build', (done) => {
 // watch changes for all files
 gulp.task('watch', () => {
   gulp.watch(allSourceFiles, ['build.ts']);
+  gulp.watch(allExampleAppfiles, ['build.ts']);
```
harunurhan: Why do we need this? @zzacharo

zzacharo: Just leftover for watching my changes on example folder. I will remove it!

zzacharo:
@harunurhan it's ok now!

harunurhan:
We should think about it again if we really need add all types of records and routing here instead of having one complicated record with no routing at all.

protractor.conf.js
```diff
@@ -7,7 +7,7 @@ var SpecReporter = require('jasmine-spec-reporter');
 exports.config = {
   allScriptsTimeout: 11000,
   specs: [
-    './e2e/**/*.e2e-spec.ts'
+    './e2e/record-types/*.e2e-spec.ts'
```
jmartinm: You are not running all the tests, only the one in this PR.

jmartinm:
The config seems to be ignored in the sample app. For example, all references are shown (instead of the 5 per page configured)

jmartinm:
> We should think about it again if we really need add all types of records and routing here instead of having one complicated record with no routing at all.

Yes, I see some pros and cons to having the sample records here. Let's discuss them and we decide the way to go.

protractor.conf.js
zzacharo: leftover sorry!

harunurhan:
Closing since we decided to have single big record instead of all of these and ignore routing.

PR #171, opened 2017-01-31
modal view better detection change

jmartinm:
* Explicitly calls markForCheck so that template gets updated after
  updating options.

* Adds new onShow callback that allows to run code after the modal is shown
  in the UI.


zzacharo:
LGTM!

PR #181, opened 2017-02-08
Expand editor when preview is hidden

jmartinm:


jmartinm:
> Do you think having id for only testing is a good thing? Although I am not sure we can't select them without id (especially middle one).

I think those columns are quite important and we might need to reference them by `id` some other time so it might be better than complicating the selectors (which might break more easily)

PR #183, opened 2017-02-08
order and alwaysShow as arrays in config

jmartinm:

### DIFF
--- a/example/app/app.config.ts
+++ b/example/app/app.config.ts
@@ -6,6 +6,14 @@ import { JsonEditorConfig } from '../../dist';
 export class AppConfig {
   jsonEditorConfig: JsonEditorConfig = {
     schemaOptions: {
+      'authors.items': {
+        order: ['full_name', 'affiliations'],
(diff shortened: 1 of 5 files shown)


src/shared/services/json-util.service.ts
```diff
@@ -27,15 +27,25 @@ export class JsonUtilService {
 
   /**
    * Returns value of the property located in dot separated path of json.
+   * Finds the appropriate key if items and properties are not in the path.
    */
   getValueInPath(json: any, path: string): any {
```
harunurhan: this actually is almost identical with https://github.com/inveniosoftware-contrib/ng2-json-editor/blob/master/src/shared/services/json-schema.service.ts#L39, can you have a look as well? If not we can maybe change it a little bit because `json-schema.service` makes more sense for this kind of utility function.

Since this utility function is already used for normal json we can leave it as it is.

jmartinm:
@harunurhan updated code, please review.

jmartinm:
PR updated.

PR #165, opened 2017-01-29
fix-issue-164: Fix for shortcut 'move up/down'

zzacharo:
(Closes #164)

Signed-off-by: Zacharias Zacharodimos <zacharias.zacharodimos@cern.ch>

PR #190, opened 2017-02-14
move to standard json-pointer format for paths

### LINKED ISSUE #186: Change from dot to slash for json path.
https://tools.ietf.org/html/rfc6901

Above is the proposed standard and already used by `jsonpatch` standard. Since we are planning to support jsonpatch (#130). Also for merging (#133) we are planning to use accept json patches for conflicts and show them in the UI to apply/reject etc. 

Therefore we should fix this issue before starting issues mentioned above.

harunurhan:
* Fixes #186

### DIFF
--- a/src/shared/services/path-util.service.spec.ts
+++ b/src/shared/services/path-util.service.spec.ts
@@ -51,4 +51,18 @@ describe('PathUtilService', () => {
     let expectedReverseIndex = 0;
     expect(pathUtilService.getElementIndexInForwardOrReversePath(path, false)).toBe(expectedReverseIndex);
   });
+
+  it('should convert path string (json-pointer) to path array', () => {
+    let pathString = '/foo/bar/0';
+    let expected = ['foo', 'bar', 0];
+    let result = pathUtilService.toPathArray(pathString);
+    expect(result).toEqual(expected);
+  });
+
+  it('should convert path array to path string (json-pointer)', () => {
+    let path = ['foo', 'bar', 0];
+    let expected = '/foo/bar/0';
+    let result = pathUtilService.toPathString(path);
+    expect(result).toEqual(expected);
+  });
 });
--- a/src/shared/services/path-util.service.ts
+++ b/src/shared/services/path-util.service.ts
@@ -26,6 +26,10 @@ import { Injectable } from '@angular/core';
 @Injectable()
 export class PathUtilService {
 
+  get separator() {
+    return '/';
+  }
+
   /**
    *
    * @param {Array<any>} path - Element's path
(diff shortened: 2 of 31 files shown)


harunurhan:
It's good to start reviewing, but should be rebased and merged after #165

src/shared/services/path-util.service.spec.ts
```diff
@@ -51,4 +51,18 @@ describe('PathUtilService', () => {
...
   });
+
+  it('should covert path string (json-pointer) to path array', () => {
+    let pathString = '/foo/bar/0';
+    let expected = ['foo', 'bar', 0];
+    let result = pathUtilService.toPathArray(pathString);
+    expect(result).toEqual(expected);
+  });
+
+  it('should convert to path array to path string (json-pointer) ', () => {
```
jmartinm:
```
should convert path array to path string (json-pointer)
```

src/shared/services/path-util.service.ts
```diff
@@ -26,6 +26,11 @@ import { Injectable } from '@angular/core';
 @Injectable()
 export class PathUtilService {
 
+  // to make sure it's constant/final
```
jmartinm: Not sure I understand the comment, is it needed?

src/shared/services/path-util.service.ts
```diff
@@ -66,5 +71,21 @@ export class PathUtilService {
     }
     return -1;
   }
+
+  /**
+   * Converts path array `/` separated path string.
```
jmartinm:
```
Converts path array to `/` separated path string.
```

Maybe we can even add an example of array and example of / separated path.

src/shared/services/path-util.service.spec.ts
```diff
@@ -51,4 +51,18 @@ describe('PathUtilService', () => {
     let expectedReverseIndex = 0;
     expect(pathUtilService.getElementIndexInForwardOrReversePath(path, false)).toBe(expectedReverseIndex);
   });
+
+  it('should covert path string (json-pointer) to path array', () => {
```
jmartinm: `covert` -> `convert `

jmartinm:
Tested it locally without problems 👍

src/shared/services/path-util.service.ts
harunurhan: Seperator string is a literal which is returned by a get method, insted of simple making it simple property (`seperator  = '/';`).

This ensures it can be never change during application lifetime which means it is a `constant`in other Java terminology `final` property. But if you think it's already clear we could remove the comment.

jmartinm: I see, was missing the Java terminology. Yes, I think we can just remove the comment.

PR #196, opened 2017-02-20
fix-issue-179: Fix slow performance issue

### LINKED ISSUE #179: Editor slowness
How to reproduce
=============

Load this record: https://cernbox.cern.ch/index.php/s/mz1M5DvSI4JSuce

And this schema: https://cernbox.cern.ch/index.php/s/A956uAU6OX5ZYII

The interface becomes very slow (specially when typing in an input). Possible problems to look at are shortcuts and too many DOM elements (inputs)

zzacharo:
* Every time that path array was needed a new object was obtained.
    So, angular was re-rendering every time all the components and
    the UI was getting slow when typing. Caching the paths fixed the
    re-rendering process only when a new path was added.
(Closes #179)

### DIFF
--- a/src/abstract-list-field/abstract-list-field.component.ts
+++ b/src/abstract-list-field/abstract-list-field.component.ts
@@ -25,6 +25,7 @@ import { List } from 'immutable';
 import { AbstractFieldComponent } from '../abstract-field';
 
 import { JsonStoreService, AppGlobalsService, TabIndexService, PathUtilService } from '../shared/services';
+import { PathCache } from '../shared/interfaces';
 
 /**
  * Abstract component to share code of common operations of all array fields
@@ -37,6 +38,7 @@ export abstract class AbstractListFieldComponent extends AbstractFieldComponent 
   values: List<any>;
   schema: Object;
   path: Array<any>;
+  pathCache: PathCache = {};
 
   constructor(public appGlobalsService: AppGlobalsService,
     public jsonStoreService: JsonStoreService,
@@ -76,7 +78,11 @@ export abstract class AbstractListFieldComponent extends AbstractFieldComponent 
    * Returns path of the property of an element at index.
    */
   getValuePath(index: number, property: string): Array<any> {
-    return this.path.concat(index, property);
+    let valuePathString = `${this.getElementPathString(index)}${this.pathUtilService.separator}${property}`;
+    if (!this.pathCache[valuePathString]) {
+      this.pathCache[valuePathString] = this.path.concat(index, property);
+    }
+    return this.pathCache[valuePathString];
   }
 
   getElementPathString(index: number): string {
--- a/src/json-editor.component.ts
+++ b/src/json-editor.component.ts
@@ -47,7 +47,7 @@ import {
   ShortcutService
 } from './shared/services';
 
-import { JsonEditorConfig, Preview, SchemaValidationErrors } from './shared/interfaces';
+import { JsonEditorConfig, Preview, SchemaValidationErrors, PathCache } from './shared/interfaces';
 
 
 @Component({
@@ -73,6 +73,7 @@ export class JsonEditorComponent extends AbstractTrackerComponent implements OnI
   previews: Array<Preview> = [];
   isPreviewerHidden: boolean;
   keys: Set<string>;
+  pathCache: PathCache = {};
 
   constructor(public http: Http,
     public appGlobalsService: AppGlobalsService,
@@ -154,7 +155,10 @@ export class JsonEditorComponent extends AbstractTrackerComponent implements OnI
   }
 
   getPathForField(field: string): Array<any> {
-    return [field];
+    if (!this.pathCache[field]) {
+      this.pathCache[field] = [field];
+    }
+    return this.pathCache[field];
   }
 
   onFieldAdd(field: string) {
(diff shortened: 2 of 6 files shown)


jmartinm:
The typing is fast indeed! :) Just testing a little bit, another thing that has a little lag is the 'Add new' button in the categories list:

slowness

Not very visible in the image, but when clicking there is a bit of delay so probably some unneeded computation happening.

@zzacharo now that you are into it, can you have a look?

zzacharo:
@jmartinm I will have a look!

src/abstract-list-field/abstract-list-field.component.ts
```diff
@@ -75,12 +76,20 @@ export abstract class AbstractListFieldComponent extends AbstractFieldComponent
...
+    let valuePathString = this.getElementPathString(index, property);
+    if (!this.paths[valuePathString]) {
+      this.paths[valuePathString] = property ? this.path.concat(index, property) : this.path.concat(index);
+    }
+    return this.paths[valuePathString];
   }
 
-  getElementPathString(index: number): string {
-    return `${this.pathString}${this.pathUtilService.separator}${index}`;
+  getElementPathString(index: number,  property?: string): string {
```
harunurhan: Please don't change this, and in the [Line#80](https://github.com/inveniosoftware-contrib/ng2-json-editor/pull/196/files#diff-0831221ed39b574c418645ef0262336eR80) you can do use it like below:

```typescript
`${this.getElementPathString(index)}${this.pathUtilService.separator}${property || ''}`
```

Because this place is the only case where we pass property to the this function.

src/abstract-list-field/abstract-list-field.component.ts
```diff
@@ -75,12 +76,20 @@ export abstract class AbstractListFieldComponent extends AbstractFieldComponent
   /**
    * Returns path of the property of an element at index.
    */
-  getValuePath(index: number, property: string): Array<any> {
-    return this.path.concat(index, property);
+  getValuePath(index: number, property?: string): Array<any> {
```
harunurhan: Actually can you override this without `property` param for `primitive-list` because looks like we're gonna have 2 conditions and it's better if we separate. In that case you don't have to use `||` in `${property || ''}` that I suggested in another comment below.

zzacharo:
@jmartinm It is lagging because when adding a new element we have here https://github.com/inveniosoftware-contrib/ng2-json-editor/blob/master/src/add-new-element-button/add-new-element-button.component.ts#L36 the focusOnElement inside setTimeout() function because we must wait for the component to be created. So that's why is giving the impression of lagging. If moved it out it will not focus on the new added element because it isn't in the DOM yet.

src/json-editor.component.ts
```diff
@@ -73,6 +73,7 @@ export class JsonEditorComponent extends AbstractTrackerComponent implements OnI
   previews: Array<Preview> = [];
   isPreviewerHidden: boolean;
   keys: Set<string>;
+  paths: {[field: string]: Array<string>} = {};
```
harunurhan: it's small but still repeated several times so maybe it's better to create a interface for this type like `PathCache` in `./shared/interfaces`

src/object-field/object-field.component.ts
```diff
@@ -43,6 +43,7 @@ export class ObjectFieldComponent extends AbstractFieldComponent implements OnIn
   @Input() path: Array<any>;
 
   keys: Set<string>;
+  paths: {[field: string]: Array<string>} = {};
```
harunurhan: ditto

src/abstract-list-field/abstract-list-field.component.ts
```diff
@@ -37,6 +37,7 @@ export abstract class AbstractListFieldComponent extends AbstractFieldComponent
   values: List<any>;
   schema: Object;
   path: Array<any>;
+  paths: {[field: string]: Array<string>} = {};
```
harunurhan: ditto

harunurhan:
@jmartinm @zzacharo `setTimeout` should never block the UI thread, I am not sure why this is happening.

zzacharo:
@harunurhan Check again!

src/json-editor.component.ts
```diff
@@ -73,6 +73,7 @@ export class JsonEditorComponent extends AbstractTrackerComponent implements OnI
   previews: Array<Preview> = [];
   isPreviewerHidden: boolean;
   keys: Set<string>;
+  paths: PathCache = {};
```
harunurhan: `pathCache` might be better I personally use plurals when the variable is an array.

src/abstract-list-field/abstract-list-field.component.ts
```diff
@@ -76,7 +78,11 @@ export abstract class AbstractListFieldComponent extends AbstractFieldComponent
    * Returns path of the property of an element at index.
    */
   getValuePath(index: number, property: string): Array<any> {
-    return this.path.concat(index, property);
+    let valuePathString = `${this.getElementPathString(index)}}${this.pathUtilService.separator}${property}`;
```
harunurhan: extra `}` in the tepmplate, right after `this.getElementPathString(index)`

src/shared/interfaces/path-cache.ts
```diff
@@ -0,0 +1,6 @@
+export interface PathCache {
+  /**
+   * Field path in string format as a key and field path in array format as a value
+   */
+  [field: string]: Array<string>;
```
harunurhan: `relativeJsonPointer` would be better instead of `field`

harunurhan:
@jmartinm LGTM, you can have the final look and merge.

Issue #187, opened 2017-02-11
Use different trackBy (by index) when it makes sense


harunurhan:
We are using a simple `trackBy` that tracks by `element`, but i some cases it might be better to use  a different one (by index perhaps).

So we must investigate performance, when a new element is appended to, removed from, moved up/down in the list. Also when a new property is added to an element of list.

Then decide where to use which kind of trackBy.

harunurhan:
Current state seems to be OK

PR #200, opened 2017-02-21
fix-issue-199: Fix tab after editing a field.

### LINKED ISSUE #199: Tab is broken when is triggered after editing a field.

zzacharo:
* Replaces trackByFunction in table-list component with
    a trackByIndex function. As a result when we update a
    specific value in a table's element, only the specifc
    element is re-rendered.
(Closes #199)


zzacharo:
related to this #187 the performance when changing the input property values of table-list.component was increased as the re-render of the components decreased.

harunurhan:
LGTM but restarted the travis build because it timed out for protractor thing.

PR #201, opened 2017-02-22
fix-issue-197: Make disabled fields untabable

zzacharo:

### DIFF
--- a/src/primitive-field/primitive-field.component.ts
+++ b/src/primitive-field/primitive-field.component.ts
@@ -105,7 +105,7 @@ export class PrimitiveFieldComponent extends AbstractFieldComponent implements O
   }
 
   get tabIndex(): number {
-    return this.tabIndexService.getElemTabIndex(this.pathString);
+    return this.schema['disabled'] ? -1 : this.tabIndexService.getElemTabIndex(this.pathString);
   }
 
   setTabIndex() {
--- a/src/shared/services/schema-fixer.service.ts
+++ b/src/shared/services/schema-fixer.service.ts
@@ -77,8 +77,12 @@ export class SchemaFixerService {
       schema = this.fixAnyOf(schema);
     } else if (schema['allOf']) {
       schema = this.fixAllOf(schema);
-    } else if (schema['order']) {
+    }
+    if (schema['order']) {
       schema = this.fixOrder(schema);
+    }
+    if (schema['disabled']) {
+      schema = this.fixDisabled(schema);
     }
     // schema fixes must be done above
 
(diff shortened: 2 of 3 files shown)


src/primitive-field/primitive-field.component.ts
```diff
@@ -105,7 +105,11 @@ export class PrimitiveFieldComponent extends AbstractFieldComponent implements O
   }
 
   get tabIndex(): number {
-    return this.tabIndexService.getElemTabIndex(this.pathString);
+    if (this.schema['disabled'] && this.schema['disabled'] === true) {
```
jmartinm: should be enough with:

```javascript
if (this.schema['disabled'] === true)
```

jmartinm:
Can you check the `Admin mode`? It does not enable the Disabled fields and I think it was doing it before, @harunurhan

zzacharo:
yes i noticed this...in master branch is working correctly?

jmartinm:
> yes i noticed this...in master branch is working correctly?

Please test, and if this PR broke it add a test to make sure it does not break in the future.

zzacharo:
No it doesn't work in master too.

src/shared/services/schema-fixer.service.ts
```diff
@@ -172,6 +192,11 @@ export class SchemaFixerService {
           };
         }
       });
+    // copy disabled attribute inside fixedSchema because it
+    // is outside anyOf element and is ignored
+    if (schema['disabled']) {
+      Object.assign(fixedSchema, {disabled: schema['disabled']});
```
harunurhan: `fixedSchema['disabled'] = true` would simply work.

src/primitive-field/primitive-field.component.ts
```diff
@@ -105,7 +105,7 @@ export class PrimitiveFieldComponent extends AbstractFieldComponent implements O
   }
 
   get tabIndex(): number {
-    return this.tabIndexService.getElemTabIndex(this.pathString);
+    return this.schema['disabled'] === true ? -1 : this.tabIndexService.getElemTabIndex(this.pathString);
```
harunurhan: `this.schema['disabled'] ? -1 : ...`

zzacharo: In theory this would also return -1 if this.schema['disabled'] is just defined.

harunurhan: `disabled` could be either `true`or `false`, so for my comment, when it's true, it'll return `-1` and when it's `false` or `undefined` it'll return real `tabindex`.

PR #202, opened 2017-02-24
tabindexes: Changes tabindex policy

zzacharo:
* Deletes tabIndexService and its references.
  * Set tabindex input of all desired tabable elements to 1. All
    elements have by default tabindex 0. If 0 was used it will tabbed
    to all elements in visual order but not only in the desired ones.
    By setting tabindex to 1 tabbing is executed in visual order is used
    and only for the elements that we want.
  * Deletes 2 e2e tests in shortcut-action.e2e.spec.ts that was checking
    for the correct update of tabindexes as now nothing is updated.


zzacharo:
@jmartinm rebased and pushed!

PR #203, opened 2017-03-02
WIP schema-validation: Schema validation errors for primitive types

zzacharo:


zzacharo:
I pushed it to check in parallel. Some questions are:
1) Do we need to categorize the error messages from validation in a more user-custom way? Right now the default error message is returned.
2) Delete in this PR the previous validation functions?

The code needs to ammend the test for schema-validation.service and probably delete the unused validation functions. Let's check if it's ok this implementation @jmartinm @harunurhan !

src/json-editor.component.html
```diff
@@ -9,7 +9,7 @@
   </div>
   <div id="editor-middle-column" [ngClass]="{widerColumn6: isPreviewerHidden}" class="col-md-6 middle" [shortcuts]="shortcuts">
     <table id="editor" class="table">
-      <div class="field-wrapper" *ngFor="let key of keys | addAlwaysShowFields:schema | filterAndSortBySchema:schema; trackBy:trackByElement">
+      <div class="field-wrapper" *ngFor="let key of keys | addAlwaysShowFields:schema | filterAndSortBySchema:schema; let last=last; trackBy:trackByElement" [ngClass]="{last:last}">
```
jmartinm: So you tried with CSS pseudo selector to apply styling to the last element instead?

src/json-editor.component.scss
```diff
@@ -128,8 +131,8 @@ td.button-holder, th.button-holder {
   width: 46px;
 }
 
-div.error {
-  background: #f2968d;
+td.error {
+  background-color: #f2968d !important;
```
jmartinm: is important needed? There is more than one `error` class inside the `td`? We can also give the class a more meaningful name to avoid adding important.

src/primitive-field/primitive-field.component.ts
```diff
@@ -103,4 +106,11 @@ export class PrimitiveFieldComponent extends AbstractFieldComponent implements O
     return this.schema['disabled'] ? -1 : 1;
   }
 
+  setTooltipPosition() {
+    let elems = document.getElementsByClassName('last');
```
jmartinm: If you are getting the elements with class `last`, why the last element would have placement `bottom` ?

src/shared/services/schema-validation.service.ts
```diff
@@ -21,17 +21,63 @@
...
+    this.ajv.addFormat('url', this.reWebUrl);
+  }
   /**
    * Validate string value against schema
    *
    * @throws {Error} if value can not be validated
    *
    * Checks: pattern
-   *
+   *s
```
jmartinm: leftover

src/json-editor.component.html
zzacharo: With the css selectors I can get all the last element but not the very last... Because the last-of-type selector selects every last element... Other than that the problem isn't to apply some css but recognize it in order to check every time on tooltip shown if the placement of the tooltip must be top or bottom...

src/json-editor.component.scss
zzacharo: No the problem is that somewhere the background color was set again... I am gonna check it again and see what causes it and fix it or let you know why important is needed.

src/primitive-field/primitive-field.component.ts
```diff
@@ -103,4 +103,15 @@ export class PrimitiveFieldComponent extends AbstractFieldComponent implements O
     return this.schema['disabled'] ? -1 : 1;
   }
 
+  get tooltipPosition() {
+    let tooltipPlacement = 'top';
+    let firstElem = document.getElementById('first');
+    if (firstElem) {
+      // The first top-level element will have the same bottom placement for all it's elements.
+      if (firstElem.querySelector(`[id='${this.pathString}']`)) {
```
harunurhan: `getElementById` ?

src/shared/services/schema-validation.service.ts
```diff
@@ -21,40 +21,71 @@
...
 
   /**
-   * Validate array value against schema
+   * Validates a specific value against schema
    *
-   * @throws {Error} if array can not be validated
+   * Uses: ajv package for json-schema validation
    *
-   * Checks: uniqueItems
+   * Returns: Object in format { isValid: boolean, error: string }
```
harunurhan: redundant since we'll have the return type.

src/shared/services/schema-validation.service.ts
```diff
@@ -21,40 +21,71 @@
...
-   *
-   */
-  validateStringValue(value: string, schema: Object) {
-    let pattern = schema['pattern'];
-    if (pattern && !value.match(pattern)) {
-      throw Error(`Does not match with pattern: ${pattern}`);
-    }
+  private ajv = new Ajv();
+  // https://gist.github.com/dperini/729294
+  private reWebUrl = new RegExp(
```
zzacharo: It covers the most cases based on a blog that I found...everything is a big word :p

src/primitive-field/primitive-field.component.ts
zzacharo: getElementById is applied to document. I don't think that it can be applied to existing element.

src/shared/services/schema-validation.service.ts
```diff
@@ -21,40 +21,71 @@
...
+   * Returns: Object in format { isValid: boolean, error: string }
    *
    */
-  validateArray(array: Array<any>, schema) {
-    if (schema['uniqueItems']) {
-      let hasUniqueItems = (new Set(array)).size === array.length;
-      if (hasUniqueItems) {
-        throw Error(`All items must be unique`);
-      }
+  validateValue(value: any, schema: Object) {
```
zzacharo: the `errors` variable is an Array<{}> as it is declared on `AbstractFieldComponent`. SchemaValidationErrors return the `json-pointer` path so it is redundant for the local validation errors as we are on the primitive component and we know the path. Actually, even to return an array is redundant because it validates a specific value and gives a specific error message.

src/primitive-field/primitive-field.component.ts
```diff
@@ -72,15 +71,16 @@ export class PrimitiveFieldComponent extends AbstractFieldComponent implements O
...
+    if (isValid) {
+      this.jsonStoreService.setIn(this.path, this.value);
+    } else {
+      errors = [{
+        message: error
+      }];
     }
-    // TODO: should we make the change even if it is not validated
-    this.jsonStoreService.setIn(this.path, this.value);
+    this.errors = errors;
```
zzacharo: If we distinguish them we must integrate them before showing the errors in the `template` in primitive-component.html.

harunurhan: well, we need to have something otherwise we loose one of them everything the other is updated.

src/primitive-field/primitive-field.component.ts
harunurhan: oh, right 👍

src/shared/services/schema-validation.service.ts
harunurhan: First of all, an error for now is `{ message: string }`, later I guess there will be `type: string` as well which could be 'warning' or 'error'. In that case error is a not a good name for this thing but I haven't found anything better so far.

`avj` returns a single string with comma separated error messages which means there could be multiple errors so that's why it is better to split it and return `Array<{ message: string }>`

zzacharo: So we should cast it to `SchemaValidationErrors` or an `Array<{}>`? The first one needs the path also.

harunurhan: I accidentally edited your comment but 
No `SchemaValidationErrors`, it's for `errorMap` only it should be `Array<{ message: string }>`

zzacharo:
@harunurhan @jmartinm check this again!

src/primitive-field/primitive-field.component.ts
jmartinm: I am not sure about this repetition, for each primitive component `document.getElementById('first')` will always match and the second query needs to be done.

I am wondering if it is better from `json-editor.component.html` to pass an `Input()` in `any-type-field` with `isLastElement`

zzacharo:
yes this is something questionable and passing an @Input() would resolve this but I don't know if it's nice. @harunurhan what do you think?

src/json-editor.component.html
```diff
@@ -9,7 +9,7 @@
   </div>
   <div id="editor-middle-column" [ngClass]="{widerColumn6: isPreviewerHidden}" class="col-md-6 middle" [shortcuts]="shortcuts">
     <table id="editor" class="table">
-      <div class="field-wrapper" *ngFor="let key of keys | addAlwaysShowFields:schema | filterAndSortBySchema:schema; trackBy:trackByElement">
+      <div class="field-wrapper" *ngFor="let key of keys | addAlwaysShowFields:schema | filterAndSortBySchema:schema; let first=first; trackBy:trackByElement" [id]="first ? 'first' : ''">
```
zzacharo: I will change the name of the `id` because using a class doing the queryselector after bad performance.

src/json-editor.component.scss
```diff
@@ -67,6 +67,10 @@ table {
     padding-right: 5px;
   }
 
+  td.error {
+    background-color: #f2968d !important;
```
zzacharo: I am gonna set it as SASS variable.

package.json
```diff
@@ -44,6 +44,7 @@
     "ts-helpers": "^1.1.1"
   },
   "dependencies": {
+    "ajv": "^4.11.3",
```
zzacharo: I will investigate this one!

src/primitive-field/primitive-field.component.ts
zzacharo: yes this is something questionable and passing an @Input() would resolve this but I don't know if it's nice. @harunurhan what do you think?

harunurhan: what we could do is put the `pathString` of first top-level property to `app-globals` then here we can check:
```typescript
if (this.pathString.startsWith(this.appGlobalsService.firstFieldPathString)
```

Naming could be changed, but this is gonna remove all queries.

@zzacharo  @jmartinm

src/json-editor.component.scss
harunurhan: I am not sure if it's OK but maybe it's better to have `.error` instead of `td.error`, I feel like we will use it more

src/primitive-field/primitive-field.component.ts
zzacharo: yes this will prevent querying all the time...I am gonna make the changes tomorrow.

jmartinm: Yep, a bit cheating with the globals, but it will work and simplify the code.

zzacharo:
@harunurhan @jmartinm check again. The only thing remaining which I am checking now is the error in the js console from ajv.

package.json
zzacharo: @jmartinm I checked that this is a warning other than error and is related with webpack. There are some proposed solutions but we must change webpack config and now is not provided from angular-cli for configuring webpack config from package.json like karma. Here is a link with the critical-dependencies issue in webpack:
https://github.com/benjamine/jsondiffpatch/issues/76

zzacharo: Also for this warning you can check here: https://github.com/webpack/webpack/issues/196

zzacharo:
@jmartinm @harunurhan I updated the PR with aligning. Check it again!

src/shared/pipes/set-first-element-path.pipe.ts
```diff
@@ -0,0 +1,39 @@
...
+
+@Pipe({
+  name: 'setFirstElementPath'
+})
+export class SetFirstElementPathPipe implements PipeTransform {
+
+  constructor(public appGlobalsService: AppGlobalsService) { }
+
+  transform(fields: Set<string>, schema: Object): Set<string> {
+    this.appGlobalsService.firstElementPath = fields.values().next().value;
```
harunurhan: `.values().next().value` to `.first()`

harunurhan: also please append `/` to the `key` so that it becomes a `path`

src/shared/services/app-globals.service.ts
```diff
@@ -29,6 +29,7 @@ import { SchemaValidationErrors } from '../interfaces';
 export class AppGlobalsService {
   private _globalErrorsSubject: ReplaySubject<SchemaValidationErrors> = new ReplaySubject<any>(1);
   public adminMode = false;
+  public firstElementPath = '';
```
harunurhan: `firstTopLevelElementPath` could more descriptive.

src/primitive-field/primitive-field.component.ts
```diff
@@ -103,4 +99,12 @@ export class PrimitiveFieldComponent extends AbstractFieldComponent implements O
     return this.schema['disabled'] ? -1 : 1;
   }
 
+  get tooltipPosition() {
+    let tooltipPlacement = 'top';
+    if (this.pathString.includes(this.appGlobalsService.firstElementPath)) {
```
harunurhan: `.startsWith` is more strict and faster.

harunurhan:
LGTM

zzacharo:
@harunurhan @jmartinm check this again to merge it before needs rebase again :). I changed only this line https://github.com/inveniosoftware-contrib/ng2-json-editor/pull/203/files#diff-d3aaa7e6a111d72c85a9914c4520ebc3R32 from string to Set because with tabs now we have more than one first element path.

zzacharo:
@harunurhan check it again.

src/shared/pipes/set-first-element-path.pipe.ts
```diff
@@ -0,0 +1,41 @@
...
+  name: 'setFirstElementPath'
+})
+export class SetFirstElementPathPipe implements PipeTransform {
+
+  constructor(public appGlobalsService: AppGlobalsService,
+              public pathUtilService: PathUtilService) { }
+
+  transform(fields: Set<string>, tabName: string): Set<string> {
+    this.appGlobalsService.firstTopLevelElementsPathMap = this.appGlobalsService.firstTopLevelElementsPathMap
+                                                            .set(tabName, `${this.pathUtilService.separator}${fields.first()}`);
```
harunurhan: formatting, just add a tab (2-spaces) after the dot.

src/shared/services/app-globals.service.ts
```diff
@@ -22,13 +22,15 @@
...
-
+import { Map } from 'immutable';
 import { SchemaValidationErrors } from '../interfaces';
 
 @Injectable()
 export class AppGlobalsService {
   private _globalErrorsSubject: ReplaySubject<SchemaValidationErrors> = new ReplaySubject<any>(1);
   public adminMode = false;
+  public activeTabName = 'Main';
+  public firstTopLevelElementsPathMap = Map<string, string>();
```
harunurhan: Doesn't have to be `Immutable.Map` or even `Map`, I think  just `{ [string: tabName]: string }` is good enough.

harunurhan: `tabNameToFirstTopLevelElement` could be a better name.

src/shared/services/app-globals.service.ts
```diff
@@ -38,4 +40,7 @@ export class AppGlobalsService {
     return this._globalErrorsSubject;
   }
 
+  get getFirstElementPathForCurrentTab() {
```
harunurhan: having get prefix in getters (`get functionName`) is redundant.

zzacharo:
@harunurhan changes done!

harunurhan:
@jmartinm  can have final look and merge it, LGTM.

However travis is failing because of some `yarn` thing?

harunurhan:
I restarted the jobs, supposedly issue fixed by `travis-ci` https://github.com/travis-ci/travis-ci/issues/7395

zzacharo:
@harunurhan perfect! Now build is successfull so we can merge I think.

PR #206, opened 2017-03-06
refactor component templates

harunurhan:
* Changes the case when a component manages two level in a path
(has two nested ngFor over the value) by re using object-field for
complex-list elements and new table-item-field for table-list elements.

* Adds list-action-group component to share template of up/down and
delete.

* Fix style which is broken because of html structure change.

### DIFF
--- a/src/list-action-group/list-action-group.component.html
+++ b/src/list-action-group/list-action-group.component.html
@@ -1,0 +1,11 @@
+<div>
+  <button type="button" class="editor-btn-delete" (click)="onDelete.emit()">
+    &times;
+  </button>
+  <button type="button" class="editor-btn-move-up" (click)="onMove.emit(-1)">
+    <i class="fa fa-chevron-up"></i>
+  </button>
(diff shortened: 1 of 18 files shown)


zzacharo:
LGTM!

harunurhan:
@jmartinm @zzacharo 
when I run it with clean `npm install`, it still works so for sure, the problem with the other travis fail is node version. It works on `6.9.0` though. So I guess it's OK to merge.

src/list-action-group/list-action-group.component.html
```diff
@@ -0,0 +1,11 @@
+<div>
+  <button type="button" class="editor-btn-delete" (click)="onDelete.emit()">
```
jmartinm: Its the fist time I see you doing this pattern. What is the reason for using `emit` instead of just calling the function?

harunurhan: What function do you mean exactly? because logic of deleting things in not inside of this component, so this is just a button group and notifies parent when a button clicked.

PR #212, opened 2017-03-20
support toggle on top

harunurhan:
## UI

screen shot 2017-03-20 at 16 12 04
screen shot 2017-03-20 at 16 13 11

@StellaCh


StellaCh:
Can we have the toggles on the right since it's easy for the users to mis-click on the toggles rather than the tabs?

harunurhan:
@StellaCh  I align them right but the css ended up a bit weird than I expected so put it in a different commit so that @jmartinm can check. If it's OK I'll squash.

PR #230, opened 2017-03-28
support item templates for autocomplete and refactor ref-field

harunurhan:
so this how it looks you can have a look (still WIP though), this is the cleanest way that I can think of, we can also use this this `templates` input for `$ref` fields in order to avoid dynamic definition and component loading.

UPDATE:

I also refactored the `ref-field` to use `<template>`, see second commit.

TODO: 
 
- [x] clean up example before the merging.


jmartinm:
Looks good to me.

harunurhan:
@jmartinm I will update it and let you know when it's ready to merge

zzacharo:
LGTM!

harunurhan:
@jmartinm cleaned up config and example app, ready to merge.

PR #231, opened 2017-03-29
add npm link documentation

harunurhan:
Today @zzacharo and I realized that this is needed 😄

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,6 +85,35 @@ npm start
 gulp watch
 ```
 
+### Working with npm link
+
+```bash
+# in ng2-json-editor
+
+# do initial build into dist
+npm run build
+
+# copy package.json to dist
+npm run copy
+
+# go to dist folder and link it
+cd dist
+npm link
+
+# remove node_modules inside dist which might cause weird errors
+rm -rf node_modules/
+
+# in another terminal session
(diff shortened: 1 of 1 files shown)


zzacharo:
:+1: Perfect :D

README.md
```diff
@@ -85,6 +85,35 @@ npm start
...
+### Working with npm link
+
+```bash
+# in ng2-json-editor
+
+# do initial build into dist
+npm run build
+
+# copy package.json to dist
+npm run copy
```
jmartinm: What was this for?

harunurhan: this is for copying `LICENSE` and `package.json`

zzacharo:
@jmartinm `package.json` is needed for linking inside `dist` folder. Otherwise, it will link from the nearest folder that includes one! @harunurhan right?

README.md
jmartinm: And why is it needed to work with `npm link` ?

zzacharo: `package.json` is needed for linking inside dist folder. Otherwise, it will link from the nearest folder that includes one.

harunurhan:
Yes, the process should be same as publishing, otherwise `npm` looks for a `package.json` in parent directories and links not only `dist`, but a lot of other stuff.

PR #239, opened 2017-04-07
fix dropdown bugs

harunurhan:
* Prevents title-dropdown from closing when find-replace is used.

* Fixes searchable-dropdown issues.


harunurhan:
@jmartinm keyboard navigation feature was also removed with the new implementation. https://github.com/valor-software/ngx-bootstrap/pull/1771#issuecomment-289063728

Issue #252, opened 2017-04-11
Create 'JSONSchema' type


harunurhan:
We have `Object` type for variables that is a schema, then we access via `['thing']`, I think we should avoid using this access unless `[variable]` to be more strongly typed. so we should create an typescript interface `JSONSchema` type with all possible keys according to draft-4. + it should include keys in `SchemaOption`.

PR #253, opened 2017-04-11
add JSONSchema type

harunurhan:
This is just experimental for #252, I changed only single file to use `JSONSchema` as type for `schema`.

It seems more clean and typing is more strict of course, what do you think? @jmartinm @zzacharo

### DIFF
--- a/src/object-field/object-field.component.html
+++ b/src/object-field/object-field.component.html
@@ -21,7 +21,7 @@
   <!-- ADD SUB FIELD FROM SCHEMA DROPDOWN -->
   <tr>
     <td class="button-holder">
-      <add-field-dropdown [fields]="keys" [pathString]="pathString" (onFieldAdd)="onFieldAdd($event)" [schema]="schema.properties">+</add-field-dropdown>
+      <add-field-dropdown [fields]="keys" [pathString]="pathString" (onFieldAdd)="onFieldAdd($event)" [schema]="schema">+</add-field-dropdown>
     </td>
     <td class="button-holder list-action-group-container">
       <!-- list-action-group (up/down and delete buttons), Set only if it's an element of an list (complex-list) -->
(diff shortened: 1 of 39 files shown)


src/shared/interfaces/json-schema.ts
```diff
@@ -0,0 +1,36 @@
+import { SchemaOption } from './schema-option';
+
+export interface JSONSchema extends SchemaOption {
```
jmartinm: Where did you get the list from? In case you missed some key, there is this example https://gist.github.com/enriched/c84a2a99f886654149908091a3183e15

Do we need to add $ref ?

src/object-field/object-field.component.html
```diff
@@ -21,7 +21,7 @@
   <!-- ADD SUB FIELD FROM SCHEMA DROPDOWN -->
   <tr>
     <td class="button-holder">
-      <add-field-dropdown [fields]="keys" [pathString]="pathString" (onFieldAdd)="onFieldAdd($event)" [schema]="schema.properties">+</add-field-dropdown>
```
jmartinm: Why is this change needed?

src/shared/interfaces/json-schema.ts
harunurhan: I got it somewhere like this link but, simplified it. My idea is to add the keys as we need/support them to keep it simple.

src/object-field/object-field.component.html
harunurhan: I also unified the `@Input() schema`, `add-field-dropdown` was taking `schema.properties` as input and that might be a bit confusing later, it's better to be able to assume that when you see `schema`, it's JSONSchema 😄

jmartinm:
I think this is a good idea. If you can fix the conflicts we can go ahead and merge it.

Issue #245, opened 2017-04-10
Left menu improvements


StellaCh:
Based on our user feedback, we discussed the following improvements:

- [x] When you select or add an item on the left menu, then the field in the editor stays highlighted. The yellow highlight goes away only on blur. One field is highlighted at a time.

- [x] Boolean toggles appear in the left menu too. When selected, the page goes up and they get also highlighted.

- [x] The tree structure of the menu items hasn't proved useful, so we keep the menu items only in the first level (delete deeper levels)

- [x] Add all fields in all tabs. When selecting a field that belongs to another tab, then the correct tab gets automatically selected

- [ ] The menu items should be in alphabetical order (otherwise it's not logical for the users)

- [x] The search bar is not needed. (ctrl-F has been proven enough for searching)

harunurhan:
@jmartinm other than deleting the code for the recursive menu, shall we have a config flag for that and make it non recursive by default?

PR #257, opened 2017-04-18
add menuMaxDepth config

harunurhan:
* Adds config to set maximum depth for tree-menu on the left,
referencing #245

### DIFF
--- a/src/tree-menu/tree-menu.component.spec.ts
+++ b/src/tree-menu/tree-menu.component.spec.ts
@@ -1,0 +1,142 @@
+/*
+ * This file is part of ng2-json-editor.
+ * Copyright (C) 2016 CERN.
+ *
+ * ng2-json-editor is free software; you can redistribute it and/or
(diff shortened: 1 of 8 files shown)


harunurhan:
@jmartinm  tests are ready

src/tree-menu/tree-menu.component.spec.ts
```diff
@@ -0,0 +1,142 @@
...
+      {
+        aString: 'bar'
+      }
+    ]
+  });
+  firstLevelKeysCount = 1;
+  allKeysCount = 5;
+}
+
+fdescribe('TreeMenuComponent', () => {
```
jmartinm: so far we used `describe`, shall we start using `fdescribe` ?

harunurhan: I changed this already, haven't pushed. This is for development to only run this spec in order to be faster :)

PR #258, opened 2017-04-18
highlights field instead of flashing

harunurhan:
* References #245


jmartinm:
There is a case not covered, when clicking on the left menu on a dropdown field (`field_categories`)

It stays always selected.

harunurhan:
@jmartinm I fixed in another commit, so that it's easier to see.

Issue #235, opened 2017-04-04
JsonSchema validation - do not complain on empty value


jmartinm:
![empty-value](https://cloud.githubusercontent.com/assets/580727/24654668/01f6eca6-193b-11e7-98f5-0a61e1c04b48.gif)

harunurhan:
Actually empty `string` shouldn't set as value, should be ignored instead.

PR #262, opened 2017-04-19
issue-235: Ignore totally empty string value

zzacharo:

### DIFF
--- a/src/primitive-field/primitive-field.component.ts
+++ b/src/primitive-field/primitive-field.component.ts
@@ -72,11 +72,13 @@ export class PrimitiveFieldComponent extends AbstractFieldComponent implements O
 
   commitValueChange() {
     // Validation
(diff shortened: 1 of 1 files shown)


harunurhan:
I think instead of ignoring empty, we should delete the key from `table-list-item` so this #235 is a bit bigger issue than we think.

zzacharo:
@harunurhan if we delete `table-list-item` the whole field will be disappeared right?

harunurhan:
@zzacharo you need to delete a key from an item of a table-list just from the record that is being emitted out. Display should be like an empty string. For I don't have a clear idea to do it but perhaps putting logic into `JsonStoreService.setIn` might be OK.

zzacharo:
@harunurhan What is needed for this to be closed?

harunurhan:
the same fix is in #318

PR #265, opened 2017-04-20
Add configurable templating for complex fields.

jmartinm:


src/complex-list-field/complex-list-field.component.ts
```diff
@@ -151,23 +149,56 @@ export class ComplexListFieldComponent extends AbstractListFieldComponent implem
...
+
+  get customTemplate(): TemplateRef<any> {
+    return this.appGlobalsService.templates[this.schema['itemTemplateName']];
+  }
+
+  get shouldDisplayTemplate(): boolean {
+    return typeof this.schema['itemTemplateName'] !== 'undefined';
+  }
+
+  private _getIndicesToDisplay(page) {
```
harunurhan: we use `_` only when we have a private variable with public `getter` and `setter`.

harunurhan: return type

harunurhan: `page: number`

src/complex-list-field/complex-list-field.component.ts
```diff
@@ -151,23 +149,56 @@ export class ComplexListFieldComponent extends AbstractListFieldComponent implem
...
-    // check if the indices includes some numbers that are out of values index range.
-    let lastIndexDiff = indices[indices.length - 1] - (this.values.size - 1);
-    if (lastIndexDiff > 0) {
-      indices.splice(indices.length - lastIndexDiff);
-    }
-    return indices;
-  }
+  getItemsForPage(page: number): Array<PaginatedItem> {
+    let items: Array<PaginatedItem> = [];
+    let indices: Array<number>;
```
harunurhan: you don't have to set the type, just move below line here and it will be OK. (init during definition)

src/complex-list-field/complex-list-field.component.ts
```diff
@@ -151,23 +149,56 @@ export class ComplexListFieldComponent extends AbstractListFieldComponent implem
...
-      .map((el, index) => start + index);
-    // check if the indices includes some numbers that are out of values index range.
-    let lastIndexDiff = indices[indices.length - 1] - (this.values.size - 1);
-    if (lastIndexDiff > 0) {
-      indices.splice(indices.length - lastIndexDiff);
-    }
-    return indices;
-  }
+  getItemsForPage(page: number): Array<PaginatedItem> {
+    let items: Array<PaginatedItem> = [];
```
harunurhan: this variable is not necessary, you could just return the value later without assigning to this.

src/complex-list-field/complex-list-field.component.ts
```diff
@@ -151,23 +149,56 @@ export class ComplexListFieldComponent extends AbstractListFieldComponent implem
...
+    let indices: Array<number>;
+    indices = this._getIndicesToDisplay(page);
 
+    items = indices.map((el) => {
+      let isEditFormVisible = false;
+      let showEditForm = this.schema['showEditForm'];
+      if (showEditForm === undefined || showEditForm(this.values.get(el))) {
```
harunurhan: `=== undefined` is not necessary in an `if` condition.

src/complex-list-field/complex-list-field.component.ts
```diff
@@ -151,23 +149,56 @@ export class ComplexListFieldComponent extends AbstractListFieldComponent implem
...
-      indices.splice(indices.length - lastIndexDiff);
-    }
-    return indices;
-  }
+  getItemsForPage(page: number): Array<PaginatedItem> {
+    let items: Array<PaginatedItem> = [];
+    let indices: Array<number>;
+    indices = this._getIndicesToDisplay(page);
 
+    items = indices.map((el) => {
```
harunurhan: maybe `el` to `index` might be better but I am not sure :)

src/complex-list-field/complex-list-field.component.ts
```diff
@@ -151,23 +149,56 @@ export class ComplexListFieldComponent extends AbstractListFieldComponent implem
...
+      let isEditFormVisible = false;
+      let showEditForm = this.schema['showEditForm'];
+      if (showEditForm === undefined || showEditForm(this.values.get(el))) {
+        isEditFormVisible = true;
+      }
+      return {
```
harunurhan: if you change `el` to `index` then you could use object literal shorthand here `{ index, isEditFormVisible }`

src/shared/interfaces/schema-option.ts
```diff
@@ -82,4 +82,19 @@ export interface SchemaOption {
    * - Toggle cannot be deleted once it is added.
    */
   toggleColor?: string;
+
```
harunurhan: since these are together, I would say create  something like`viewTemplateConfig` which has these two config inside`

src/shared/interfaces/schema-option.ts
```diff
@@ -82,4 +82,19 @@ export interface SchemaOption {
...
+  /**
+   * Used for complex fields only (object with nested objects inside).
```
harunurhan: complex list fields only (list of nested objects)

src/shared/interfaces/schema-option.ts
```diff
@@ -82,4 +82,19 @@ export interface SchemaOption {
...
+   */
+  itemTemplateName?: string;
+
+  /**
+   * Used for complex fields only (object with nested objects inside).
+   * Function that receives an immutable Map representing the JSON document and returns:
+   * - true if the fields to edit should be visible by default
+   * - false if the fields to edit should be hidden by default
+   */
+  showEditForm?: any;
```
harunurhan: function signature should be specified.

src/complex-list-field/complex-list-field.component.html
```diff
@@ -39,11 +39,19 @@
...
-  <div *ngFor="let pIndex of paginatedIndices; let i = index; trackBy:trackByIndex">
+  <div *ngFor="let item of paginatedItems; let i = index; trackBy:trackByIndex">
     <div class="complex-list-field-wrapper">
-      <object-field [value]="values.get(pIndex)" [schema]="schema.items" [path]="getElementPath(pIndex)">
-        <list-action-group (onMove)="moveElement(i, $event)" (onDelete)="deleteElement(i)"></list-action-group>
-      </object-field>
+      <span *ngIf="shouldDisplayTemplate && values.get(item.index).keySeq().size != 0">
+        <ng-template [ngTemplateOutlet]="customTemplate" [ngTemplateOutletContext]="{item: values.get(item.index)}"></ng-template>
+        <a href="#" (click)="item.isEditFormVisible = !item.isEditFormVisible">
+          {{item.isEditFormVisible? 'Hide Fields' : 'Show Fields'}}
```
harunurhan: space after `?`

src/complex-list-field/complex-list-field.component.ts
jmartinm: Yep, I was not sure since `map` can take 2 arguments `(el, index)` and in this case what we are using is the element (which is an index from the `indices` variable). But yeah, even if it could be confusing, I think `index` is better.

src/complex-list-field/complex-list-field.component.html
jmartinm: You mean before?

zzacharo: after `item.isEditFormVisible` before `?` :P

harunurhan: yes before 😄 , sorry about that!

jmartinm:
@harunurhan you can have another look to all the parts.

example/app/app.config.ts
```diff
@@ -38,8 +40,14 @@ export class AppConfig {
...
             return JSON.stringify(value).search(expression) > -1;
           },
-          itemsPerPage: 5,
+          itemsPerPage: 20,
           maxVisiblePageCount: 5
+        },
+        viewTemplate: {
+          itemTemplateName: 'referenceTemplate',
+          showEditForm: (value) => {
+            return !(value.get('record') && value.get('record').get('$ref'));
```
harunurhan: you could use only `value.getIn(['record', '$ref']) without having first check.

example/app/app.component.scss
```diff
@@ -0,0 +1,14 @@
+.fa-check-circle-success {
+  color: green;
+}
+
+.fa-exclamation-triangle-warning {
```
harunurhan: you can use `.fa-exclamation-triangle.warning` instead, then html class will be shorter,
`fa-exclamation-triangle fa-exclamation-triangle-warning` to `fa-exclamation-triangle warning`

harunurhan: Also for other class.

example/app/app.component.ts
```diff
@@ -34,11 +34,24 @@ import { AppConfig } from './app.config';
     'app.component.scss'
   ],
   template: `
+    <ng-template let-item="item" #referenceTemplate>
```
harunurhan: could we even do something simpler? or in the case since this template is too big to embed in ts, you can create `app.component.html` instead.

example/app/app.config.ts
```diff
@@ -38,8 +40,14 @@ export class AppConfig {
           findMultiple: (value, expression) => {
             return JSON.stringify(value).search(expression) > -1;
           },
-          itemsPerPage: 5,
+          itemsPerPage: 20,
           maxVisiblePageCount: 5
+        },
+        viewTemplate: {
```
harunurhan: this won't compile, should be `viewTemplateConfig`

src/shared/interfaces/view-template-config.ts
```diff
@@ -0,0 +1,17 @@
+import { Map } from 'immutable';
+
+export interface ViewTemplateConfig {
+    /**
+     * Used for complex list fields only (list of nested objects).
+     * The template needs to be defined with <ng-template #myTemplate></ng-template> and passed to the json-editor like so:
+     * <json-editor [templates]="{myTemplate: myTemplate}"></json-editor>
```
harunurhan: can you add 
```markdown
- template parameter: `item: Immutable.Map<string, any>`
```

src/complex-list-field/complex-list-field.component.ts
```diff
@@ -151,23 +144,49 @@ export class ComplexListFieldComponent extends AbstractListFieldComponent implem
...
-    }
-    return indices;
-  }
+  getItemsForPage(page: number): Array<PaginatedItem> {
+    let indices = this.getIndicesToDisplay(page);
 
+    return indices.map((index) => {
+      let isEditFormVisible = false;
+      let showEditForm = this.schema.viewTemplateConfig ? this.schema.viewTemplateConfig.showEditForm : undefined;
+      if (!showEditForm || showEditForm(this.values.get(index))) {
```
harunurhan: `let isEditFormVisible = !showEditForm || showEditForm(this.values.get(index))` would be better.

jmartinm:
Updated. Had to add a click handler (https://github.com/inveniosoftware-contrib/ng2-json-editor/pull/265/files#diff-c1ece59e02735f4e0f0470f4006229c8R46) to prevent default since I tried it in the `record-editor` and otherwise the page was reloading.

example/app/app.component.scss
```diff
@@ -0,0 +1,14 @@
+.fa-check-circle.success {
+  color: green;
+}
+
+.fa-exclamation-triangle.warning {
+  color: orange;
+}
+
+.referenceTemplateContainer {
```
harunurhan: `kebab-case`

example/app/app.config.ts
```diff
@@ -1,5 +1,7 @@
 import { Injectable } from '@angular/core';
 
+import { Map } from 'immutable';
```
harunurhan: is this necessary?

jmartinm:
@harunurhan just rebased this, once tests pass should be good to go.

PR #266, opened 2017-04-21
Update mock data to latest INSPIRE schema

jmartinm:


harunurhan:
Great!, you could merge it right away when travis passes :), also careful about the example config!

PR #268, opened 2017-04-24
@Output naming styling

jmartinm:
* Renames @Output to follow convention from
  https://angular.io/docs/ts/latest/guide/style-guide.html#!#05-16


david-caro:
Looks old, is this still relevant?

PR #273, opened 2017-04-24
refactor pathCache and fix #271

### LINKED ISSUE #271: pathCache is not being invalidated properly
Usages of `pathCache` - like https://github.com/inveniosoftware-contrib/ng2-json-editor/blob/master/src/object-field/object-field.component.ts#L47 are not being properly invalidated.

How to reproduce
=============

In the references field, add a new one and check the value of `record` in the root component. The reference is not updated properly.

harunurhan:
* Refactors pathCache and moves getting path for child logic to
abstract-field.component so that every component shares.

* Fixes #271

### DIFF
--- a/src/abstract-field/abstract-field.component.ts
+++ b/src/abstract-field/abstract-field.component.ts
@@ -25,7 +25,7 @@ import { Subscription } from 'rxjs/Subscription';
 
 import { AbstractTrackerComponent } from '../abstract-tracker';
 import { AppGlobalsService, PathUtilService } from '../shared/services';
-import { ValidationError} from '../shared/interfaces';
+import { ValidationError, PathCache} from '../shared/interfaces';
 
 /**
  * This is the base class for fields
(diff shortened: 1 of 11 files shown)


src/abstract-field/abstract-field.component.ts
```diff
@@ -62,6 +63,19 @@ export abstract class AbstractFieldComponent
...
   }
 
+  /**
+   * Gets path for child, returns from `pathCache` if it is a hit
+   * in order not to re-render child component due to reference change its path.
+   *
+   * @param key - index or field name for child
+   */
+  getPathForChild(key: any): Array<any> {
+    if (!this.pathCache[key] || this.pathCache[key][this.path.length - 1] !== this.path[this.path.length - 1]) {
```
harunurhan: `this.pathCache[key][this.path.length - 1] !== this.path[this.path.length - 1]` part is the trick that fixes the issue here, others are refactoring.

jmartinm:
Since we have decided to merge as is, just rebase to fix the conflict.

PR #283, opened 2017-04-27
Style changes

jmartinm:


src/editor-previewer/editor-previewer.component.scss
```diff
@@ -1,7 +1,7 @@
 div.preview-container {
   width: 100%;
+  height: 95vh;
```
harunurhan: I think we should use `height: 100%;` instead. If you think it's better to do later, we should create an issue then.

jmartinm: Yes, we can do later because I tried the other day and it was not that simple. From the parent div column it seems that you need to set `height:100%` to all elements all the way down until this `preview-container`.

PR #284, opened 2017-04-28
Fix tab keys missing

jmartinm:


src/shared/services/tabs-util.service.ts
```diff
@@ -70,6 +71,17 @@ export class TabsUtilService {
     return keyToTabName;
   }
 
+  getTabNameToKeys(keys: Set<string>, schemaKeyToTabName: { [key: string]: string } , tabNames: Array<string>):
+  { [tabName: string]: Set<string> } {
+    let tabNameToKeys = keys.groupBy(key => schemaKeyToTabName[key]).toObject() as any;
+    for (let tabName of tabNames) {
```
harunurhan: We don't use `for...of` but use `.forEach`. Well there is no special reason but you can convert this `for` to something like this.

```typescript
tabNames
  .filter(tabName => !(tabName in tabNameToKeys))
  .forEach(tabName =>{ tabNameToKeys[tabName] = Set() })
```

src/shared/services/tabs-util.service.ts
```diff
@@ -70,6 +71,17 @@ export class TabsUtilService {
     return keyToTabName;
   }
 
+  getTabNameToKeys(keys: Set<string>, schemaKeyToTabName: { [key: string]: string } , tabNames: Array<string>):
```
harunurhan: no `space` after before `comma`, I thought that was enforced by `tslint` but apparently not.

src/shared/services/tabs-util.service.ts
```diff
@@ -70,6 +71,15 @@ export class TabsUtilService {
     return keyToTabName;
   }
 
+  getTabNameToKeys(keys: Set<string>, schemaKeyToTabName: { [key: string]: string }, tabNames: Array<string>):
```
harunurhan: not much logic but still a simple test for this would be better

PR #295, opened 2017-05-12
Keys store

### LINKED ISSUE #227: Make alwaysShow field undeletable and exclude from add-field-dropdown

harunurhan:
- Closes #295 
- Closes #227


src/shared/services/keys-store.service.ts
```diff
@@ -0,0 +1,162 @@
...
+import { Injectable } from '@angular/core';
+import { Map, List, OrderedSet, Iterable, Set, Seq } from 'immutable';
+import { ReplaySubject } from 'rxjs/ReplaySubject';
+
+import { JSONSchema } from '../interfaces';
+import { PathUtilService } from './path-util.service';
+
+@Injectable()
+export class KeysStoreService {
+  private keys$Map: { [path: string]: ReplaySubject<OrderedSet<string>> };
```
harunurhan: Since the `keys` wich are in the map, are `Subject` here no the map itself, I added suffix to the `keys`

src/json-editor.component.ts
```diff
@@ -138,22 +139,15 @@ export class JsonEditorComponent extends AbstractTrackerComponent implements OnI
...
     // setup variables need for tab grouping.
     if (this.config.tabsConfig) {
-      this.schemaKeyToTabName = this.tabsUtilService.getSchemaKeyToTabName(this.config.tabsConfig, this.schema);
       this.tabNames = this.tabsUtilService.getTabNames(this.config.tabsConfig);
-      this.tabNameToKeys = this.tabsUtilService.getTabNameToKeys(this.keys, this.schemaKeyToTabName, this.tabNames);
-      this.tabNameToSubSchema = this.tabsUtilService.getTabNameToSubSchema(this.schema, this.schemaKeyToTabName);
-      this.tabsUtilService.tabSelectionSubject.subscribe(tabName => {
-        this.activeTabName = tabName;
-      });
+      this.tabsUtilService.tabSelectionSubject.subscribe(tabName => { this.appGlobalsService.activeTabName = tabName; });
```
harunurhan: I agree, even removing the subject would be better since `$` suffix already means that it's an observable kind of thing.

src/shared/services/keys-store.service.ts
```diff
@@ -0,0 +1,162 @@
...
+        .forEach(key => {
+          let nextPath = `${path}${this.pathUtilService.separator}${key}`;
+          this.buildKeysMapRecursively(map.get(key), schema.properties[key], nextPath);
+        });
+    } else if (schema.componentType === 'table-list') {
+      let list = mapOrList as List<Map<string, any>>;
+      this.buildKeysForTableList(path, list, schema);
+      // there is no recursive call for table list items because they aren't expected to have object or object list as property.
+      // FIXME: really?
+    } else {
```
harunurhan: the thing is, this fucntion is never called for a primitive type if you checked the recursive calls. So this else is actually executed for `complex-list`

PR #298, opened 2017-05-17
fix-issue-289: Add shortcut for adding below item

zzacharo:

### DIFF
--- a/e2e/shortcut-action/shortcut-action.e2e-spec.ts
+++ b/e2e/shortcut-action/shortcut-action.e2e-spec.ts
@@ -57,6 +57,17 @@ describe('ShortcutAction', function () {
       });
   });
 
+  it(`should add a new row under table using 'mod+shift+b' shortcut`, () => {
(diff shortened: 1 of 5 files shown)


zzacharo:
@harunurhan I updated the interface

jmartinm:
`mod+shift+b` does not work on Mac. The shortcut is already taken.

zzacharo:
@jmartinm mac again :P Which should I choose?

jmartinm:
https://support.google.com/chrome/answer/157179?hl=en

`mod+shift+c` seems to be free :P

zzacharo:
I think there is a problem with that because I used it on copy from root and we changed it!

jmartinm:
> I think there is a problem with that because I used it on copy from root and we changed it!

Oh, I will try locally and let you know

jmartinm:
Nevermind. Actually `mod+shift+b` is working fine (I just had not added your commit from this PR :P). The shortcut seems to take precedence over the Chrome one, so it works.

PR #299, opened 2017-05-17
add linkTextBuilder for ref-field

### LINKED ISSUE #293: Replace all urls with icons
Testers argued that there's too much noise in the interface from urls that are not useful to see.

Wherever we have the field 'record', replace it with an icon that leads to the link described:
link

Like we have for 
url

harunurhan:
* Fixes #293

* Adds example config (could remove if wanted @jmartinm)

* Small bug fixes for ref-field

### DIFF
--- a/src/ref-field/ref-field.component.ts
+++ b/src/ref-field/ref-field.component.ts
@@ -35,7 +35,7 @@ import { Observable } from 'rxjs/Observable';
 import 'rxjs/add/operator/catch';
 
 import { AppGlobalsService, PathUtilService } from '../shared/services';
-import { RefConfig, JSONSchema } from '../shared/interfaces';
+import { RefConfig, JSONSchema, RefAnchorAttributes } from '../shared/interfaces';
 
 @Component({
   selector: 'ref-field',
@@ -56,6 +56,8 @@ export class RefFieldComponent implements OnChanges {
 
   isPreviewButtonHidden = false;
 
+  private anchorAttributes: RefAnchorAttributes;
+
   constructor(private http: Http,
     private changeDetectorRef: ChangeDetectorRef,
     private appGlobalsService: AppGlobalsService,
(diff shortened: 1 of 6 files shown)


src/ref-field/ref-field.component.ts
```diff
@@ -112,7 +124,7 @@ export class RefFieldComponent implements OnChanges {
       .get(this.ref, this.requestOptions)
       .map(res => res.json())
       .catch(error => {
-        return Observable.of({ error });
+        return Observable.of({ $error: error });
```
jmartinm: In the other PRs you had the dollar sign at the end `error$`

jmartinm:
With this change, even though the link will now say: `View conference` the link will point to the JSON representation. I think the catalogers will find more useful for the link to go the main page of the conference, so `/conference/1234`

jmartinm:
> With this change, even though the link will now say: View conference the link will point to the JSON representation. I think the catalogers will find more useful for the link to go the main page of the conference, so /conference/1234

So it might be more flexible to have a config `linkTemplateName` that will receive the `$ref` value and can add arbitrary HTML so that we can personalize the link, wdyt?

harunurhan:
@jmartinm  we could also change the config function to return an object of `{ link: string, text: string }` so that it's more flexible?

harunurhan:
@jmartinm  changes made, you could have a look again.

src/shared/interfaces/ref-config.ts
```diff
@@ -1,21 +1,28 @@
...
-  lazy: boolean;
+  lazy?: boolean;
 
   /**
    * Array of headers which will be used for http request that fetches the data in `$ref`
    */
   headers?: Array<HttpHeader>;
+
+  /**
+   * Function to be called to set display and href of the acnhor element that opens the $ref link on the new tab
```
jmartinm: `anchor

src/shared/interfaces/ref-config.ts
```diff
@@ -1,21 +1,28 @@
 import { HttpHeader } from './http-header';
+import { RefAnchorAttributes } from './ref-anchor-attributes';
 
 export interface RefConfig {
   /**
    * Name of the template where you can access the data which is fetched from `$ref` url.
    *
    * - `@Input() templates` for `<json-editor>` must have an entry where key is this and value is a `TemplateRef`
-   * - template parameter: `response: Object` is json body, `response.error` is used to propagate http error during fetchin `$ref`
+   * - template parameter: `response: Object` is json body, `response.$error` is used to propagate http error during fetchin `$ref`
```
jmartinm: `fetching`

PR #302, opened 2017-05-19
issue-243: Fixes admin bug to reflect changes automatically

zzacharo:
* Fixes automatic change detection on disabled fields

### DIFF
--- a/src/any-type-field/any-type-field.component.ts
+++ b/src/any-type-field/any-type-field.component.ts
@@ -24,6 +24,7 @@ import {
   Component,
   Input,
   ChangeDetectionStrategy,
+  ChangeDetectorRef
 } from '@angular/core';
 
 import { AppGlobalsService, ComponentTypeService, PathUtilService } from '../shared/services';
@@ -54,7 +55,13 @@ export class AnyTypeFieldComponent {
   @Input() path: Array<any>;
   @Input() value: any;
 
-  constructor(public appGlobalsService: AppGlobalsService) { }
+
+  constructor(public changeDetectorRef: ChangeDetectorRef,
+    public appGlobalsService: AppGlobalsService) {
+    this.appGlobalsService.adminModeSubject.subscribe(adminMode => {
+      this.changeDetectorRef.markForCheck();
+    });
+  }
 
   get isDisabled(): boolean {
     return this.schema && this.schema.disabled && !this.appGlobalsService.adminMode;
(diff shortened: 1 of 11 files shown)


zzacharo:
Needs fixing for hidden fields

zzacharo:
Is ready!

src/sub-record/sub-record.component.ts
```diff
@@ -57,8 +57,12 @@ export class SubRecordComponent extends AbstractTrackerComponent implements OnCh
   constructor(public jsonStoreService: JsonStoreService,
     public appGlobalsService: AppGlobalsService,
     public tabsUtilService: TabsUtilService,
+    public changeDetectorRef: ChangeDetectorRef,
     public keysStoreService: KeysStoreService) {
     super();
+    this.appGlobalsService.adminModeSubject.subscribe(adminMode => {
```
jmartinm: Is this needed? Since you already `markForCheck` in primitive-field-component (which is a child of this component) this should be checked as well.

src/object-field/object-field.component.ts
```diff
@@ -49,6 +49,9 @@ export class ObjectFieldComponent extends AbstractFieldComponent {
     public changeDetectorRef: ChangeDetectorRef,
     public keysStoreService: KeysStoreService) {
     super(appGlobalsService, pathUtilService, changeDetectorRef);
+    this.appGlobalsService.adminModeSubject.subscribe(adminMode => {
```
jmartinm: Is this needed? Since you already markForCheck in primitive-field-component (which is a child of this component) this should be checked as well.

src/object-field/object-field.component.ts
```diff
@@ -61,4 +64,7 @@ export class ObjectFieldComponent extends AbstractFieldComponent {
     this.keysStoreService.deleteKey(this.pathString, name);
   }
 
+  isDisabled(key): boolean {
+    return this.schema.properties[key].disabled && !this.appGlobalsService.adminMode ;
```
jmartinm: extra space at the end.

src/any-type-field/any-type-field.component.ts
```diff
@@ -54,7 +55,13 @@ export class AnyTypeFieldComponent {
   @Input() path: Array<any>;
   @Input() value: any;
 
-  constructor(public appGlobalsService: AppGlobalsService) { }
+
+  constructor(public changeDetectorRef: ChangeDetectorRef,
+    public appGlobalsService: AppGlobalsService) {
+    this.appGlobalsService.adminModeSubject.subscribe(adminMode => {
```
jmartinm: Is this needed? Since you already markForCheck in primitive-field-component (which is a child of this component) this should be checked as well.

harunurhan:
#305

PR #303, opened 2017-05-23
table css styling

jmartinm:
* Improves contrast in between fields and minor fixes.


jmartinm:
@harunurhan please install it locally and check that everything looks ok.

PR #308, opened 2017-05-24
Add katex support

jmartinm:


jmartinm:
@harunurhan updated this so you can start reviewing it.

Still `katex.service.ts` can be simplified a bit.

jmartinm:
Now pushed with `katex.service.ts` simplified.

src/string-input/string-input.component.ts
```diff
@@ -0,0 +1,83 @@
...
+  ],
+  templateUrl: './string-input.component.html',
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class StringInputComponent implements OnInit {
+
+  @Input() value: string;
+  @Input() path: Array<any>;
+  @Input() pathString: string;
+  @Input() title: string;
```
harunurhan: should be called `placeholder`

src/string-input/string-input.component.ts
```diff
@@ -0,0 +1,83 @@
...
+  @Input() tabIndex: number;
+  @Input() latexPreviewEnabled: boolean;
+
+  @Output() blur = new EventEmitter<void>();
+  @Output() keyPress = new EventEmitter<void>();
```
harunurhan: `keypress` would be better since the native one is like this.

src/string-input/string-input.component.ts
```diff
@@ -0,0 +1,83 @@
...
+  styleUrls: [
+    './string-input.component.scss'
+  ],
+  templateUrl: './string-input.component.html',
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class StringInputComponent implements OnInit {
+
+  @Input() value: string;
+  @Input() path: Array<any>;
```
harunurhan: not used probably

src/string-input/string-input.component.ts
```diff
@@ -0,0 +1,83 @@
...
+  @Input() pathString: string;
```
harunurhan: also not necessary if since `getElementById` could be avoided by using`@ViewChild` decorator. 


https://angular.io/docs/ts/latest/api/core/index/ViewChild-decorator.html

src/shared/services/katex.service.ts
```diff
@@ -0,0 +1,168 @@
...
+        }
+        fragment.appendChild(span);
+      }
+    }
+
+    el.replaceChild(fragment, el.childNodes[0]);
+
+  }
+
+  findEndOfMath(delimiter: string, text: string, startIndex: number): number {
```
harunurhan: `private`

src/shared/services/katex.service.ts
```diff
@@ -0,0 +1,168 @@
...
+import { render, ParseError } from 'katex';
+
+import { JSONSchema } from '../interfaces';
+
+@Injectable()
+export class KatexService {
+  private delimiters = [
+    { left: '$$', right: '$$', display: true },
+    { left: '$', right: '$', display: false },
+    // FIXME check if we need the ones below
```
harunurhan: ?

src/shared/services/katex.service.ts
```diff
@@ -0,0 +1,168 @@
...
+    { left: '\\[', right: '\\]', display: true },
+    { left: '\\(', right: '\\)', display: false },
+  ];
+
+  renderMathInText(text: string, el: HTMLElement) {
```
harunurhan: this one needs some documentation, so that we know it renders latex text into a given html element.

src/shared/services/katex.service.ts
jmartinm: I am just not sure in all our records all the ways math formulas can appear. They can be in between $ or $$ and also I have seen in between `\begin{equation}` and `\end{equation}`

jmartinm:
Updated.

PR #314, opened 2017-05-31
fix deleting list element bugs

### LINKED ISSUE #310: Error div in a complex field makes it not possible to be removed
screen shot 2017-05-29 at 1 11 06 pm

Delete button is not clickable in the `doi` field.

harunurhan:
* Fixes #310


jmartinm:
By checking this locally I have seen a couple of weird things (might not be related)

1) If you add to a reference field a `doi` and then delete the `doi`, a new error is added to the error panel even though there is no value.

2) Errors are not shown when tabbing outside of a field. If I am inside of the `doi` field, type a wrong one and tab, the error is not shown.

PR #316, opened 2017-05-31
fix searchable-dropdown autoclose bug

### LINKED ISSUE #300: Dropdown closes automatically if an input has been edited before
How to reproduce
=============

1) Edit an input
2) Click on a field with a dropdown.

The dropdown will briefly open and close automatically.

categories

harunurhan:
* Fixes #300


harunurhan:
Good thing that we haven't merged, I found an issue and fixed the issue completely with a different thing (by handling show-hide logic manually). At some point I'd like to consider moving away from ng2-bootstrap for this dropdowns because it's not reliable and testable, needs always workarounds.

@jmartinm

PR #321, opened 2017-06-07
fix-issue-315: Fixes error panel update on field deletion

zzacharo:
* Fixes tabbing outside a field by marking it for check.

### DIFF
--- a/src/abstract-field/abstract-field.component.ts
+++ b/src/abstract-field/abstract-field.component.ts
@@ -36,14 +36,13 @@ import { ValidationError, PathCache} from '../shared/interfaces';
  * It provides trackByFunction from AbstractTrackerComponent, and handles errors for the component.
  */
 export abstract class AbstractFieldComponent
-  extends AbstractTrackerComponent
-  implements OnInit, OnDestroy {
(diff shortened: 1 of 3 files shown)


src/abstract-field/abstract-field.component.ts
```diff
@@ -76,15 +70,12 @@ export abstract class AbstractFieldComponent
...
 
-  get isErrorTooltipEnabled(): boolean {
-    return this.hasErrors;
-  }
-
   get pathString(): string {
     return this.pathUtilService.toPathString(this.path);
   }
 
-  get hasErrors() {
```
harunurhan: this hould be removed but should be changed to

`return this.externalErrors.length > 0`

src/primitive-field/primitive-field.component.ts
```diff
@@ -117,4 +124,23 @@ export class PrimitiveFieldComponent extends AbstractFieldComponent {
     return this.disabled ? 'disabled' : '';
   }
 
+  get hasErrors(): boolean {
+    return this.externalErrors.length > 0 || this.internalErrors.length > 0;
```
harunurhan: `return super.hasErrors || this.internalErrors.length > 0;`

PR #324, opened 2017-06-20
validation: Set custom date format validation

zzacharo:

### DIFF
--- a/src/shared/services/schema-validation.service.ts
+++ b/src/shared/services/schema-validation.service.ts
@@ -71,9 +71,27 @@ export class SchemaValidationService {
     '$', 'i'
   );
 
+  private validateDate(value) {
+    let dateFormats = [
+      '^\\d{4}$',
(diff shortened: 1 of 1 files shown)


src/shared/services/schema-validation.service.ts
```diff
@@ -71,9 +71,27 @@ export class SchemaValidationService {
     '$', 'i'
   );
 
+  private validateDate(value) {
+    let dateFormats = [
```
jmartinm: Since this is quite specific of `record-editor`, is it possible to move this customization of the validation to `record-editor` via, e.g some config?

jmartinm:
I will merge this since it is better than what we have and create a separate issue for moving to `record-editor`

src/shared/services/schema-validation.service.ts
zzacharo: yes absolutely!

PR #335, opened 2017-07-27
add non immutable support to json-store.service

harunurhan:
* Now a javascript object or array can be passed to addIn and setIn
functions, and they are later converted to Immutable. Thanks to this
parent app doesn't need to care about complexity of immutable data
structures.


src/shared/services/json-store.service.ts
```diff
@@ -23,6 +23,8 @@ export class JsonStoreService {
       return;
     }
 
+    value = this.convertToImmutableIfNot(value);
```
jmartinm: My only comment would be to just rename this function to `toImmutable()`

PR #338, opened 2017-08-07
Merge ui

harunurhan:


src/abstract-field/abstract-field.component.ts
```diff
@@ -42,23 +42,35 @@ export abstract class AbstractFieldComponent
...
+      .map(patchesByPath => patchesByPath[this.pathString])
+      .subscribe(patches => {
+        this.jsonPatches = patches || [];
+        this.removeJsonPatch = this.jsonPatches
+          .find(patch => patch.op === 'remove');
+      });
   }
   /**
-   * Gets path for child, returns from `pathCache` if it is a hit
+   * Gets path for child, returns from `pathCache` if it is a hit'''''''////////////////////,
```
harunurhan: leftover

src/patch-actions/patch-actions.component.ts
```diff
@@ -0,0 +1,54 @@
+/*
+ * This file is part of ng2-json-editor.
+ * Copyright (C) 2016 CERN.
```
harunurhan: 2017

src/primitive-field/primitive-field.component.ts
```diff
@@ -134,7 +134,7 @@ export class PrimitiveFieldComponent extends AbstractFieldComponent implements O
   }
 
   get tabIndex(): number {
-    return this.schema.disabled ? -1 : 1;
+    return this.disabled ? -1 : 1;
```
harunurhan: ?

src/primitive-field/primitive-field.component.html
```diff
@@ -1,51 +1,62 @@
...
+            <li *ngFor="let error of internalErrors">
+              {{error.message}}
+            </li>
+            <li *ngFor="let error of externalErrors">
+              {{error.message}}
+            </li>
+          </ul>
+        </ng-template>
+        <td *ngIf="!jsonPatches[0]; else patchTemplate" class="value-container" [ngClass]="disabledClass" [tooltip]="errorsTooltipTemplate"
+          [isDisabled]="!hasErrors()" placement="{{tooltipPosition}}" container="body">
```
harunurhan: hasError?

src/patch-view/patch-view.component.html
```diff
@@ -0,0 +1,47 @@
+<div class="patch-container green-left-border" [id]="patch.path" tabindex="-1">
```
harunurhan: should be named `add-patch-view`

src/shared/services/json-store.service.ts
```diff
@@ -60,8 +65,16 @@ export class JsonStoreService {
...
       value = this.toImmutable(value);
-      list = lastPathElement === '-' ? list.push(value) : list.insert(lastPathElement, value);
+      if (lastPathElement === '-') {
+        list = list.push(value);
+        path[path.length - 1] = list.size - 1;
+      } else {
+        list =  list.insert(lastPathElement, value);
+      }
       this.setIn(pathWithoutIndex, list);
+      if (Map.isMap(value)) {
```
harunurhan: is this checked also inside of `buildKeysMapRecursivelyForPath`

jmartinm:
I have tried locally and these are some problems I have found:

* Error when deleting the `/references/0/reference/misc/0` field.
* Delete the whole field abstracts, then when creating a new one the previous conflict shows.
* Accepting the remove patch in `/references/3` crashes
* Might be unrelated to this PR, but accept the patch in `/references/-`, then move it up and two things: the fields get hidden by default which should not be and when shown they have extra keys.
* Adding a `remove` patch in any keyword field does not work (gives an error approving it)
* If a reference has a patch and then we move that reference up and down the patches get applied to the wrong reference.
* Changing page in references applies the patch to the wrong reference. If the second item of first page has a patch, then the second item of the second page will get the patch. Also it does not seem possible to add a patch to an element of the second page.

PR #355, opened 2017-08-17
swap keys while moving complex list items

### LINKED ISSUE #354: KeysStore should be updated while moving elements

harunurhan:
* Fixes #354

* Adds moveIn utility to json-store so that moving will be shared btw
components and shortcuts.

* Adds test for deleteKey.

* Removes types from method comments.

### DIFF
--- a/src/shared/services/json-store.service.ts
+++ b/src/shared/services/json-store.service.ts
@@ -47,7 +47,7 @@ export class JsonStoreService {
     return this.json.getIn(path);
   }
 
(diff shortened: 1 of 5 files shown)


src/shared/services/json-store.service.ts
```diff
@@ -81,6 +81,30 @@ export class JsonStoreService {
     }
   }
 
+  /**
+   * Moves the element at given index P or DOWN within the list
```
harunurhan: `P` to `UP`

PR #359, opened 2017-08-21
add switch to see only items with edit form

harunurhan:
* Adds switch to see only complex-list items to edit.

* Changes .active class color for complex-list navigator.

## UI

screen shot 2017-08-21 at 16 42 07


harunurhan:
Only thing that can be done in this PR or left for later is to blocking/hiding  UP/DOWN buttons and shortcuts.

PR #361, opened 2017-08-21
support undo for setting new value to top level

### LINKED ISSUE #358: Push to History when setting top-level fields
We should push the revert-patch to the history when setting top-level fields.

harunurhan:
* Fixes #358

* Fixes undo crash while trying to remove undo patch fron global
json-patches which are associated to components.

### DIFF
--- a/src/shared/services/json-store.service.ts
+++ b/src/shared/services/json-store.service.ts
@@ -21,7 +21,7 @@ export class JsonStoreService {
   constructor(private pathUtilService: PathUtilService,
     private keysStoreService: KeysStoreService) { }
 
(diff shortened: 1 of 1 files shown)


src/shared/services/json-store.service.ts
```diff
@@ -39,6 +39,17 @@ export class JsonStoreService {
         this.json = this.json.setIn(pathToIndex, List());
       }
     }
+
+    // save revert patch for undo if it's all document or top-level
+    if (path.length <= 1) {
+      this.history.push({
```
harunurhan: don't push to history if this is called by `rollbackLastChange`

PR #364, opened 2017-08-22
fix focusing to latex fields (closes #336)

### LINKED ISSUE #336: doesn't focus latex fields

harunurhan:


harunurhan:
@jmartinm could be merged, e2e will be added after #366 is resolved.

PR #375, opened 2017-08-31
fix undo crash

harunurhan:
* Fixes undo problem but only works when an array item is deleted not
when a property is deleted.


jmartinm:
Why doesnt it work when a property is deleted ?

harunurhan:
@jmartinm 
Deleted property should be added to its keys of its parent back.

But `json-store` and `keys-store` are turning a bit like spaghetti that's why I didn't wanna do a quick fix. I will think about it a bit more and possibly find a better solution.

PR #425, opened 2017-10-25
add patch actions to patches console tab

### LINKED ISSUE #422: accept/reject all json-patches from bottom console

harunurhan:
* Closes #422 

### Initial UI
screen shot 2017-10-25 at 15 20 48


StellaCh:
Looks ok, I'd suggest to make the buttons smaller than the tabs. In this step, the buttons are not the primary user action, they appear just as a support for the user, hence no need to take so much space in the interface and distract the user. Right now, it looks like the user is supposed to click one of the Accept All or Reject All buttons and that's not the case.

PR #456, opened 2018-01-23
make the sidebar collapsible

Dinika:
The sidebar collapses to a narrower bar, like so:
screenshot from 2018-02-12 11-20-18

On a smaller screen (iPad):
screenshot from 2018-02-12 11-27-37

screenshot from 2018-02-12 11-28-30

When the sidebar is expanded:
screenshot from 2018-02-12 11-29-17

### DIFF
--- a/src/add-field-dropdown/add-field-dropdown.component.html
+++ b/src/add-field-dropdown/add-field-dropdown.component.html
@@ -1,6 +1,6 @@
 <div *ngIf="!hidden" class="btn-group add-field-dropdown-container" dropdown keyboardNav="true" [isDisabled]="isDisabled" (onShown)="onDropdownShown()">
   <button type="button" class="btn btn-add-field-dropdown" dropdownToggle>
-    <ng-content></ng-content> <i class="fa fa-caret-down"></i>
+     <ng-content></ng-content> 
   </button>
   <ul class="dropdown-menu" *dropdownMenu>
     <li class="dropdown-filter-container">
--- a/src/add-field-dropdown/add-field-dropdown.component.scss
+++ b/src/add-field-dropdown/add-field-dropdown.component.scss
@@ -19,7 +19,6 @@
 
 .dropdown-filter-container {
   padding: 0 3px;
-  
   input {
     width: 100%;
   }
--- a/src/bottom-console-badges/bottom-console-badges.component.html
+++ b/src/bottom-console-badges/bottom-console-badges.component.html
@@ -1,11 +1,17 @@
 <div class="badges-container">
-  <a *ngIf="errorCount > 0" class="error" (click)="onBadgeClick($event, 'Errors')">
-        <span class="">Errors</span> <span class="badge">{{errorCount}}</span>
-    </a>
-  <a *ngIf="warningCount > 0" class="warning" (click)="onBadgeClick($event, 'Warnings')">
-        <span>Warnings</span> <span class="badge">{{warningCount}}</span>
+  <a *ngIf="errorCount > 0" class="error" (click)="onBadgeClick($event, 'Errors')" tooltip="{{errorCount}} error(s)" placement="right">
+    <span class="error fa-stack fa-lg">
+      <i class="fa fa-times fa-stack-1x fa-inverse"></i>
+    </span>
   </a>
-  <a *ngIf="patchCount > 0" class="patch" (click)="onBadgeClick($event, 'Patches')">
-    <span>Conflicts</span> <span class="badge">{{patchCount}}</span>
+  <a *ngIf="warningCount > 0" class="warning" (click)="onBadgeClick($event, 'Warnings')" tooltip="{{warningCount}} warning(s)" placement="right">
+    <span class="warning fa-stack fa-lg"> 
+      <i class="fa fa-exclamation fa-stack-1x fa-inverse"></i>
+    </span>
+  </a>
+  <a *ngIf="patchCount > 0" class="patch" (click)="onBadgeClick($event, 'Patches')" tooltip="{{patchCount}} conflicts(s)" placement="right">
+    <span class="patch fa-stack fa-lg">
+      <i class="fa fa-bolt fa-stack-1x fa-inverse"></i>
+    </span>
   </a>
 </div>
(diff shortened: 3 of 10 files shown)


StellaCh:
Comparing the 2 options, I believe that the icons should have an outline, as in the last screenshot, as in that way it's more clear that they are buttons that can be pushed. The error panel area is something that we definitely want the user to pay attention to, and it helps it they are more dominant in the page

harunurhan:
@StellaCh @Dinika how about displaying the number of errors, warnings etc.?

In my opinion, it is nice to see them without opening the console, or at least icon color could be grayish or something if there is no error, warning etc.

src/json-editor.component.ts
```diff
@@ -290,4 +277,10 @@ export class JsonEditorComponent extends AbstractSubscriberComponent implements
     return element;
   }
 
+  toggleSidebar() {
+    this.isSidebarCollapsed = !this.isSidebarCollapsed;
+    document.querySelector('#expanded-sidebar').classList.toggle('open');
```
harunurhan: We never use native api to handle ui unless it's really necessary. This doesn't seem to be one of those cases. The ideal solution for this to make a boolean state (component property) which you already have actually `isSidebarCollapsed`

Then to apply a class on html tag you can bind the boolean

`[class.open]="!isSidebarCollapsed"`

there are also other ways to do it, if you need to apply multiple classes etc.

https://angular.io/api/common/NgClass

src/json-editor.component.scss
```diff
@@ -184,10 +195,105 @@ body, html {
...
+        transition: all 0.5s ease-in-out;
+        -webkit-transition: all 0.5s ease-in-out;
+        -ms-transition: all 0.5s ease-in-out;
+        &:hover {
+          color: white;
+          text-shadow: 3px 3px 14px #2a5d88;
+        }
+      }
+    }
+    #open-sidebar {
```
harunurhan: using `id` to apply styles only acceptable if you have to do really hacky thing which is discouraged :).

if you need to this in a `e2e` test, you might use `id` but write another class to appy the styles.

(`#ng2-json-editor` is an exception, since it's root selector used to scope our global styles)

drjova:
@harunurhan AFAIR the icons are displayed only when there is an actual number. IMHO @Dinika's suggestion, to display the number with a tooltip, will work well.

harunurhan:
Another thing for the UI design, I think when the menu is collapsed, the main part is too wide, AFAIK users were requesting a wider preview. Probably @StellaCh can provide more precise info.

StellaCh:
That's right. In a discussion IRL we discussed that the extra space should be shared between the editor and the preview so that the pdf preview takes more space.

src/add-field-dropdown/add-field-dropdown.component.html
```diff
@@ -1,7 +1,5 @@
 <div *ngIf="!hidden" class="btn-group add-field-dropdown-container" dropdown keyboardNav="true" [isDisabled]="isDisabled" (onShown)="onDropdownShown()">
-  <button type="button" class="btn btn-add-field-dropdown" dropdownToggle>
-    <ng-content></ng-content> <i class="fa fa-caret-down"></i>
-  </button>
+  <i class="fa fa-plus fa-lg add-field-button" aria-hidden="true" dropdownToggle tooltip="Add Field" placement="right"></i>
```
harunurhan: this will break an untested part, so maybe we can also introduce a test after fixing.

src/add-field-dropdown/add-field-dropdown.component.scss
```diff
@@ -1,26 +1,26 @@
...
-  text-shadow: 0 1px 0 #fff;
-  color: darkslategray;
-  margin-bottom: 2px;
-  float: left;
-
-  &:hover {
-    color: blue;
-    opacity: 0.6;
-  }
+.dropdown-menu {
```
harunurhan: all files 2 spaces tab.

src/bottom-console-badges/bottom-console-badges.component.html
```diff
@@ -1,11 +1,23 @@
 <div class="badges-container">
-  <a *ngIf="errorCount > 0" class="error" (click)="onBadgeClick($event, 'Errors')">
-        <span class="">Errors</span> <span class="badge">{{errorCount}}</span>
-    </a>
-  <a *ngIf="warningCount > 0" class="warning" (click)="onBadgeClick($event, 'Warnings')">
-        <span>Warnings</span> <span class="badge">{{warningCount}}</span>
+  <a *ngIf="errorCount > 0" class="error" (click)="onBadgeClick($event, 'Errors')"
```
harunurhan: again formatting is bad
```html
<foo>
  <bar>
     <thing>A thing</thing>
  </bar>
</foo>
```

src/json-editor.component.scss
```diff
@@ -76,19 +82,27 @@ $label-holder-button-color: #595959;
   padding: 0px;
 }
 
-html, body {
+body, html {
```
harunurhan: unnecessary

src/json-editor.component.ts
```diff
@@ -30,7 +30,7 @@ import {
   SimpleChanges,
   ViewEncapsulation,
   ChangeDetectionStrategy,
-  TemplateRef
+  TemplateRef,
```
harunurhan: unnecessary.

src/add-field-dropdown/add-field-dropdown.component.html
harunurhan: Later I saw that, this change and other scss are breaking the UI of all other cases we use `add-field-dropdown`

harunurhan:
@Dinika can you just squash them into one or if it makes sense, couple of commits.

Dinika:
@harunurhan Just completed the squash.

Dinika:
@drjova I updated the PR to fix the alignment of the badge icons. Do you think this looks better?
screenshot from 2018-02-12 11-20-30

Dinika:
file:///home/dinika/Pictures/Screenshot%20from%202018-02-12%2011-20-30.png

Dinika:
Do you think this looks better?

screenshot from 2018-02-12 11-20-30

harunurhan:
@drjova can be merged unless you need UI change

drjova:
@Dinika looks great thanks!

PR #458, opened 2018-01-24
remove html, body style

### LINKED ISSUE #457: remove legacy `body`, `html` styles

harunurhan:
* Fixes #457


harunurhan:
Strangely, we rely on `height:100%;`, so it had to stay until we fix the style issue.

PR #463, opened 2018-02-12
add icon in sidebar to toggle preview

Dinika:
Screnshots:
1. Maximized Preview
screenshot from 2018-02-14 14-11-05

2. Minimized Preview
screenshot from 2018-02-14 14-13-52

Signed-off-by: Dinika Saxena <dinika.saxena@cern.ch>

### DIFF
--- a/e2e/preview-panel/preview-panel.e2e-spec.ts
+++ b/e2e/preview-panel/preview-panel.e2e-spec.ts
@@ -33,16 +33,14 @@ describe('Preview panel', function() {
   });
 
   it(`should expand/collapse preview window`, () => {
-    const previewToggleButton = element(by.id('btn-preview-toggle'));
+    const previewToggleIcon = element(by.id('preview-toggle-icon'));
     const middleColumn = element(by.id('middle-main-container'));
     const rightColumn = element(by.id('right-main-container'));
-
-    expect(previewToggleButton.getText()).toEqual('Hide Preview');
 
     const middleColumnWidthBefore = middleColumn.getCssValue('width');
     const rightColumnWidthBefore = rightColumn.getCssValue('width');
 
-    previewToggleButton.click();
+    previewToggleIcon.click();
 
     const middleColumnWidthAfter = middleColumn.getCssValue('width');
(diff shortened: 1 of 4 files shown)


src/json-editor.component.html
```diff
@@ -22,7 +27,7 @@
   </div>
 
   <div id="middle-main-container" class="middle main-container" [class.compact]="config.compact" [shortcuts]="customShortcutKeys"
-    [ngClass]="isPreviewerHidden ? 'maximizeEditor' : 'maximizeEditor'">
+    [ngClass]="isPreviewerHidden ? 'maximizeEditor' : '' ">
```
harunurhan: since you are fixing this it's better to use single class to boolean style biding. After some renaming something like  `[class.editor-fullscreen]="isPreviewHidden"` would be better.

e2e/preview-panel/preview-panel.e2e-spec.ts
```diff
@@ -37,8 +37,6 @@ describe('Preview panel', function() {
     const middleColumn = element(by.id('middle-main-container'));
     const rightColumn = element(by.id('right-main-container'));
 
-    expect(previewToggleButton.getText()).toEqual('Hide Preview');
```
harunurhan: also you need remove the line where `previewToggleButton` is selected

Dinika: I think I would have to keep this line because later on the test "clicks" the `previewToggleIcon` (which is the id now given to a span element instead of a button) to check that the sizes of the containers change after the click.

e2e/preview-panel/preview-panel.e2e-spec.ts
```diff
@@ -33,16 +33,14 @@ describe('Preview panel', function() {
...
-
     const middleColumnWidthBefore = middleColumn.getCssValue('width');
     const rightColumnWidthBefore = rightColumn.getCssValue('width');
 
-    previewToggleButton.click();
+    previewToggleIcon.click();
```
Dinika: @harunhuran This is where the `previewToggleIcon` is being used. (I've just changed the name here)

PR #475, opened 2018-04-26
services: test case for ``undefined`` pathces

drjova:
Signed-off-by: Harris Tzovanakis <me@drjova.com>


harunurhan:
Quick fix for the test, would be adding empty array as default param to `setJsonPatches`.

PR #485, opened 2018-12-17
Previewer option from config

topless:
* `isPreviewerHidden` now gets a default value `false` from the config file.

### DIFF
--- a/src/shared/interfaces/json-editor-config.ts
+++ b/src/shared/interfaces/json-editor-config.ts
@@ -47,6 +47,13 @@ export interface JsonEditorConfig {
   schemaOptions?: SchemaOptions;
 
   /**
+   * If set, it hides the preview section of the editor.
+   *
+   * - It is `false` by default.
+   */
+  isPreviewerHiddenOnStart?: boolean;
+
+  /**
    * Configuration for previews to be displayed in previewer (on the right side).
    */
   previews?: Array<PreviewConfig>;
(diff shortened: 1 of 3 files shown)


example/configs/config.ts
```diff
@@ -147,6 +147,7 @@ export const defaultConfig: JsonEditorConfig = {
       }
     }
   },
+  isPreviewerHidden: false,
```
harunurhan: you can remove this, since it's already default.

src/shared/interfaces/json-editor-config.ts
```diff
@@ -46,6 +46,13 @@ export interface JsonEditorConfig {
    */
   schemaOptions?: SchemaOptions;
 
+  /**
+   * If set, it hides the preview section of the editor.
+   *
+   * - It is `false` by default.
+   */
+  isPreviewerHidden?: boolean;
```
harunurhan: can you rename it to `isPreviewerHiddenOnStart` ?

topless: I kept the property name that is used from the ui when shows/hides the previewer. There is also another option `isPreviewDisabled` so I tried not to create yet another one. But I agree is more self explained.

harunurhan: Yeah, since it's not always hidden and can be toggled later, I'd go with explicit name.

PR #491, opened 2019-02-06
hidden fields are not shown in the add dropdown

zzacharo:


src/add-field-dropdown/add-field-dropdown.component.ts
```diff
@@ -48,6 +49,10 @@ export class AddFieldDropdownComponent implements OnChanges {
     private domUtilService: DomUtilService,
     private keysStoreService: KeysStoreService) { }
 
+  get schemaProperties() {
+    return _.pickBy(this.schema.properties, value => !value.hidden);
```
harunurhan: what would be better is to remove the `differentKeys` pipe from the template, then do
```typescript
this.addableKeys = Set.fromKeys(schema.properties)
  .substract(this.fields)
  .filter(field => !schema.properties[field].hidden) 
```

on `ngOnChanges`

src/add-field-dropdown/add-field-dropdown.component.ts
```diff
@@ -43,6 +44,7 @@ export class AddFieldDropdownComponent implements OnChanges {
 
   expression = '';
   hidden = false;
+  addableKeys = <any>[];
```
harunurhan: `= Set()`

PR #508, opened 2020-10-21
ui: fix sticky tabs

oguzdemirbasci:
Ref: #1436

### DIFF
--- a/src/complex-list-field/complex-list-field.component.scss
+++ b/src/complex-list-field/complex-list-field.component.scss
@@ -7,7 +7,7 @@
 .top-bar-container {
   width: 100%;
   position: sticky;
-  top: 0;
+  top: 46px;
   z-index: 1;
   display: flex;
   align-items: center;


src/complex-list-field/complex-list-field.component.scss
```diff
@@ -7,7 +7,7 @@
 .top-bar-container {
   width: 100%;
   position: sticky;
-  top: 0;
+  top: 46px;
```
harunurhan, line 10: this looks like an hack, is `width` of the tab bar fixed to `46px` ?

oguzdemirbasci: I guess you mean `height` of tab bar. No it is not fixed directly, but it has 46px height with font size, paddings and margins. We need to set `top` as the height of the previos component for multiple sticky components.

harunurhan: If they have the same `z` they should be already stacking together, at least this how I imagine layout works in the browser.

oguzdemirbasci: As far as know, when they have the same z-index, browser renders them in the implementation order without stacking since they both have `sticky` and `top:0` properties. So, search bar hides the navbar when we scroll down.

Yes, I completely agree with you. But, to stack multiple sticky components, we need to define top values (first one is 0, second one is the height of first one and so on). To solve the issue that you mentioned, we can dynamically set the top values using JS which is not a better solution I think.

PR #509, opened 2020-10-29
Fix Sticky Tabs

oguzdemirbasci:
Ref: #1436

### DIFF
--- a/src/complex-list-field/complex-list-field.component.scss
+++ b/src/complex-list-field/complex-list-field.component.scss
@@ -7,7 +7,7 @@
 .top-bar-container {
   width: 100%;
   position: sticky;
-  top: 0;
+  top: 46px;
   z-index: 1;
   display: flex;
   align-items: center;


src/complex-list-field/complex-list-field.component.scss
```diff
@@ -7,7 +7,7 @@
 .top-bar-container {
   width: 100%;
   position: sticky;
-  top: 0;
+  top: 46px;
```
drjova, line 10: why `46px`

oguzdemirbasci: https://inspirebeta.net/editor/holdingpen/12181

I want to stick the search bar just below the nav-bar in the references tab. So we should set it to the height of nav-bar,

drjova (approved):

harunurhan:
Why was this done? I don't even see this commit now on `master` anyway

PR #510, opened 2020-11-03
fix auto-complete field selection by clicking bug

harunurhan:
For some reason underlying ngx-bootstrap's typeahead wasn't working
consistently for selecting a completion by mouse click.

This is simplest and fastest solution without touching the
ngx-bootstrap's code.

NOTE: upgrading ngx-bootstrap wasn't an option since we
have to run angular 4 and and ngx-bootstrap 2.


drjova (approved):
LGTM 👏 I guess we don't have tests in this library to verify that everything works.

harunurhan:
@drjova idea is that we have e2e for some autocomplete on `record-editor`

PR #511, opened 2020-11-10
ref: addition of string-input

drjova:
* Ref inspirehep/inspirehep#1255

### DIFF
--- a/src/ref-field/ref-field.component.html
+++ b/src/ref-field/ref-field.component.html
@@ -1,5 +1,8 @@
 <div [id]="pathString">
   <div *ngIf="ref">
+    <div *ngIf="shouldDisplayInputField" >
+      <string-input [pathString]="pathStringWithRef" [value]="ref" [tabIndex]="tabIndex" [placeholder]="schema.properties.$ref.title"></string-input>
+    </div>
     <div *ngIf="isTemplateEnabled">
       <button class="btn-preview-template" *ngIf="!shouldDisplayTemplate" (click)="onPreviewClick($event)"><i class="fa fa-eye"></i></button>
       <div *ngIf="shouldDisplayTemplate">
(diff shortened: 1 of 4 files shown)


src/ref-field/ref-field.component.html
```diff
@@ -8,7 +8,8 @@
       </div>
     </div>
     <div *ngIf="!isTemplateEnabled">
-      <a target="_blank" class="break-word" [href]="anchorHref">{{anchorDisplay}}</a>
+      <string-input *ngIf="shouldDisplayInputField" [pathString]="pathString" [value]="ref" [tabIndex]="tabIndex" [placeholder]="schema.title"></string-input>
```
harunurhan, line 11: I think this shouldn't be under `*ngIf="!isTemplateEnabled"`, we might want to display template (currently not used) even if `displayInputField` is enabled.

harunurhan: `pathString` for this should be `<pathString>/$ref`

harunurhan: Depending on the schema and what makes sense, maybe it's better to use `schema.properties.$ref.title` instead as placeholder.

harunurhan: `[value]` should be actual `$ref` value, not hardcoded string.

harunurhan (changes requested):

src/ref-field/ref-field.component.ts
```diff
@@ -100,6 +100,10 @@ export class RefFieldComponent implements OnChanges {
     return this.appGlobalsService.templates[this.refConfig.templateName];
   }
 
+  get pathStringWithRef(): string {
+    return this.pathString + '/$ref';
```
harunurhan, line 104: You should use `PathUtilService` for this.

drjova: interesting 😄

harunurhan (approved):

PR #512, opened 2020-11-16
ref: fix input field display

drjova:


src/ref-field/ref-field.component.ts
```diff
@@ -100,8 +100,8 @@ export class RefFieldComponent implements OnChanges {
     return this.appGlobalsService.templates[this.refConfig.templateName];
   }
 
-  get pathStringWithRef(): string {
-    return this.pathString + this.pathUtilService.separator + '$ref';
+  get pathArrayWithRef(): Array<any> {
+    return this.pathUtilService.toPathArray(this.pathString + this.pathUtilService.separator + '$ref');
```
harunurhan, line 104: `return this.path.concat('$ref')` is simpler

harunurhan: or you can keep and update it like that.
```js
 if (changes['path']) {
  ...
  this.refPath = this.path.concat('$ref');
}
```

PR #525, opened 2021-12-01
validate also after value change

MJedr:
ref:inspirehep/inspirehep#2269


src/primitive-field/primitive-field.component.ts
```diff
@@ -114,6 +114,7 @@ export class PrimitiveFieldComponent extends AbstractFieldComponent implements O
         this.schema.onValueChange(this.path, this.value, this.jsonStoreService, this.keysStoreService)
```
drjova, line 114: I would add it here, only if there is `onValueChange`

michamos:
just to make sure: is this automatically stripping whitespace at the beginning and end?

MJedr:
no, it would be done in editor in inspirehep repo

michamos:
ok, just making sure there's no misunderstanding :)

drjova (approved):

